PR #4, opened 2014-12-15
Add Maven pom.xml for one of the samples

dsyer:
There's a lot of (apparently quite unecessary) configuration just to
get Maven to look for sources in src/documentation/*. Also there's
duplication of the build/generated-documentation path in
documentation.properties as well as asciidoctor-maven-plugin config.
It's unidiomatic to have to configure that kind of thing in a properties
file anyway for maven, so probably better to allow it to be specified
as a restdocs plugin parameter (when such a thing exists).

But, hey, it works! And the documentation looks pretty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .classpath
 .gradle
 .project
-.settings
-bin
-build
+.settings/
+bin/
+build/
+target/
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for this, Dave. I've reworked things so that they're no longer quite so Gradle-specific. See 6ec522069a2dca8c5ae95a11839afe8636ca09db for details. Both samples can now be built with Maven and, thankfully, there's less configuration required.

PR #22, opened 2015-02-22
Removed references to the spring-restdocs-core dependency and changed them into spring-restdocs.

andreasevers:
The spring-restdocs-core dependency is not used anymore but can still be retrieved from https://repo.spring.io/snapshot.
To avoid confusion it's important the readme file is up-to-date with the right dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@ plugins {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Well spotted. Thanks for the PR.

Issue #27, opened 2015-03-03
Allow specifying prefix URI to be included in curl for loadbalancers


MikeBra:
Our curl url looks something like this
$ curl https://mycloud/router/users
The controller code only maps "/users" because the router is external to the service.

I am looking for rest documentation configurer to support a "withPrefixUri" command like the following so that the curl commands are actually valid.
new RestDocumentationConfigurer().withScheme("https").withHost("mycloud")
                        .withPort(443).withPrefixUri("/router")

p.s. ingenious idea to let the tests drive the documentation. Developers can now write tests & documentation first before implementation - Test & Doc Driven Development. (TDDD)

wilkinsona:
I'm not sure that there's a generally applicable way to solve this. If the service was using hypermedia then all of the links would also require the same modification to their URIs, and any code in the service that extracts ids etc from those links would have to understand the modification.

Depending on how your app is written, and whether or not it uses hypermedia, the best solution may be to change the dispatcher servlet's mapping. Perhaps you can share some more information about your app so that we can explore that approach?

MikeBra:
The front end router takes requests and puts them on a AMQP message exchagne for the correct tenant. We then have a service that takes messages off the queue and routes them to the correct RequestMapping service. From the perspective of the service it doesn't know about the front end router that is doing the dispatching to the proper tenant so we can't address this by adding it to the mapping. 

Adding the suggested optional prefix seems like the most flexible solution. Since it is optional it wouldn't impact existing users but would support more sophisticated routing of requests in a multi-tenant architecture.

wilkinsona:
Thanks, @mikebra.

> Adding the suggested optional prefix seems like the most flexible solution. Since it is optional it wouldn't impact existing users but would support more sophisticated routing of requests in a multi-tenant architecture.

My concern isn't so much with it impacting existing users but with the solution being (somewhat) generally applicable. For example, I'd like it to work for anyone who's documenting a hypermedia-based API that's behind a proxy/router.

My current thinking is that using the `X-Forwarded-Prefix` header may be appropriate here. The header could be set via `withPrefixUri` or similar and then used when creating the curl url. It would also have to be ignored when determining the curl command's header arguments. Setting the header would allow any link-creation logic to generate the correct URI ([and would get this for free if it's using Spring Framework's ServletUriComponentsBuilder](https://jira.spring.io/browse/SPR-12500)).

wilkinsona:
[...]

PR #49, opened 2015-04-16
Feature/context path configuration

mayboroda:
Hi, we are using spring-restdocs on our project and we are really missing `contextPath` configuration in the `RestDocumentationConfigurer`. Here is my pull-request. I think this pull request resolves the issue https://github.com/spring-projects/spring-restdocs/issues/27
Thanks in advance.


mayboroda:
Hey, sorry for not being accurate, but I think this pull request resolves the issue https://github.com/spring-projects/spring-restdocs/issues/27

wilkinsona:
Excellent. Thanks very much.

PR #52, opened 2015-04-18
Improve quoting for cURL

dewet22:
Single quotes are generally safer to use since shells won't try to interpolate
any strings or do variable substitution inside it. This could be refined to do
a check for [&%$] inside the URI/query string before just blindly surrounding
with quotes, but this is safe as it is. I'd also argue it improve readability to not
have to deal with mentally unescaping the command-line.

Also, don't escape POST paylods since they are quoted already -- this doesn't
work correctly right now.


dewet22:
Ugh, I didn't want github to merge the two commits into one pull request. Let me know if you want me to split them out or not.

wilkinsona:
Thanks for the PR. Given that the changes are unrelated, would you mind splitting them out please? Also, have you haven't already done so, can you please sign the [CLA](https://support.springsource.com/spring_committer_signup)?

dewet22:
Certainly, I'll submit two new PRs.

Issue #46, opened 2015-03-26
Add support for documenting fields in XML payloads


yohanliyanage:
Do you plan to support XML payloads as well?

wilkinsona:
For some aspects of the documentation there's nothing to be done to support XML, for example a cURL request snippet for POSTing an XML payload. On the other hand, the current support for documenting links would require you to write your own `LinkExtractor` as the two built-in extractors are both JSON-specific.

Can you be a bit more specific, what sort of XML payload support are you looking for?

yohanliyanage:
Thanks for the quick response. I am trying to add XML based examples in addition to the JSON based examples to my documentation. I'm not doing HATEOS, so I believe LinkExtractor is not relevant in my case.

When I try to extract response fields from the payload as follows, it seems like the FieldValidator is expecting JSON instead of XML.

Unit Test Code:
	this.mockMvc.perform(get("/api/v1/users").accept(MediaType.APPLICATION_XML))
	    .andExpect(status().isOk())
	    .andDo(document("users-list-example")
	            .withResponseFields(
	                    fieldWithPath("content").description("An array of <<resources-user, User resources>>"),
	                    fieldWithPath("total").description("Total records found"),
	                    fieldWithPath("pageSize").description("Size of a single page"),
	                    fieldWithPath("page").description("Current page number (zero based)"),
	                    fieldWithPath("pages").description("Total number of pages")
	            ));

Test Failure:
    com.fasterxml.jackson.core.JsonParseException: Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
     at [Source: java.io.StringReader@6621a4ef; line: 1, column: 2]
        at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1419)
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:508)
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:437)
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleOddValue(ReaderBasedJsonParser.java:1462)
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:683)
        at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3105)
[...]

PR #54, opened 2015-04-19
Add handler for custom-writing JSON objects

dewet22:
When tests query live servers, we often cannot rely on them sending nicely-formatted
JSON responses, nor can we alter their configuration. This change allows for injecting
a Jackson ObjectMapper and ObjectWriter in order to allow rewriting the received JSON
response payloads into something more documentation-friendly.

### DIFF
--- a/spring-restdocs/src/main/java/org/springframework/restdocs/RestDocumentationResultHandler.java
+++ b/spring-restdocs/src/main/java/org/springframework/restdocs/RestDocumentationResultHandler.java
@@ -19,6 +19,7 @@ package org.springframework.restdocs;
 import static org.springframework.restdocs.curl.CurlDocumentation.documentCurlRequest;
 import static org.springframework.restdocs.http.HttpDocumentation.documentHttpRequest;
 import static org.springframework.restdocs.http.HttpDocumentation.documentHttpResponse;
+import static org.springframework.restdocs.http.HttpDocumentation.documentHttpResponseWithPrettyJson;
 import static org.springframework.restdocs.hypermedia.HypermediaDocumentation.documentLinks;
 import static org.springframework.restdocs.payload.PayloadDocumentation.documentRequestFields;
 import static org.springframework.restdocs.payload.PayloadDocumentation.documentResponseFields;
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR. I like the idea, but I'm not totally sold on the implementation.

There have already been requests for better support for XML payloads (see #46, for example) and I'd prefer not to add to the problem of things being too JSON-centric.

How about a `ResponseFormatter` abstraction that takes the response as input and returns the formatted content? The default implementation, `IdentityResponseFormatter` would simply return the content as-is. To enable pretty-printing, or any other sort of response post-processing, users could associate a `ResponseFormatter` implementation with a particular content type, perhaps via an API on `RestDocumentationConfigurer` that sets the formatters on `RestDocumentationContext`. Jackson-based implementations for pretty-printing JSON and XML could be provided.

The above-described changes would probably mean that there's only a single snippet produced, rather than one snippet for the formatted response and one for the unformatted response. I suspect that won't be a problem, though, as I can't imagine someone wanting to use both the formatted and the unformatted snippets in the same documentation. That's only a hunch. If you have real word experience the says otherwise, please do say so.

dewet22:
I've been thinking about this too, and I agree. We even have a need where we want to filter out headers from the response (for instance, it isn't useful documenting things like Cache-Control, Pragma or X-Frame-Options in every request). So a ResponseFormatter would be a useful abstraction where we can fine-tune exactly how headers are filtered and response bodies are formatted.

I agree that a single snippet is fine -- I was hoping to do that but because `documentHttpResponse` gets added to the list of delegates in the constructor, it is harder to fine-tune. Do you think the default set of delegates should be configurable? Or maybe this point becomes moot when we can inject our own ResponseFormatter.

Let me take another look at this and get back to you.

wilkinsona:
Interesting. I hadn't considered stripping out headers. I like that idea too. Given that capability, perhaps `ResponsePostProcessor` would be more appropriate? It's no longer just about formatting.

I haven't looked in detail to confirm that this'll work, but I think that `RestDocumentationResultHandler` could call the configured response formatters/post-processors and then call its delegates. There's no setter for an `MvcResult`'s response so I think the result would have to be copied to contain the new response. There may also be some hoops to jump through to produce a response with formatted content: judicious use of `setCommitted(false)` and `resetBuffer()` should do the trick.

wilkinsona:
@dewet22 Thanks again for opening this PR and for inspiring a large chunk of what I've implemented in e5f35c9. Those changes should be available in the latest snapshots. If you get a chance to try them out, I'd be interested to hear any feedback you may have.

dewet22:
Thanks, I have been playing with it since I saw the commit -- it is looking very good right now. As soon as I've converted over all of our current documenting actions, I'll get back with any feedback.

PR #55, opened 2015-04-19
Improve quoting for cURL

dewet22:
Single quotes are generally safer to use since shells won't try to interpolate
any strings or do variable substitution inside it. This could be refined to do
a check for [&%$] inside the URI/query string before just blindly surrounding
with quotes, but this is safe as it is. I'd also argue it improve readability to not
have to deal with mentally unescaping the command-line.

Also, don't escape POST payloads since they are quoted already -- this doesn't
work correctly right now.


wilkinsona:
Thanks

Issue #56, opened 2015-04-19
Content-Type with charset is not supported


openwms:
Having a request with a Content-Type header field set to 'application/json;charset=UTF-8' does not match a proper LinkExtractor. I expect to get an AtomLinkExtractor for all application/json content types.

PR #57, opened 2015-04-19
Whenever the first part of the Content-Type header field matches, a Link...

openwms:
...Extractor is returned. CharSet information at the end of the Content-Type has no affect. Fixes gh-56

### DIFF
--- a/spring-restdocs/src/main/java/org/springframework/restdocs/hypermedia/LinkExtractors.java
+++ b/spring-restdocs/src/main/java/org/springframework/restdocs/hypermedia/LinkExtractors.java
@@ -17,17 +17,12 @@
 package org.springframework.restdocs.hypermedia;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.http.MediaType;
 import org.springframework.mock.web.MockHttpServletResponse;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Static factory methods providing a selection of {@link LinkExtractor link extractors}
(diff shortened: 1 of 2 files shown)


AGoliath:
@openwms  you beat me to it, I was about to raise the same issue (haven´t prepared a pr yet, though). I don´t know if `startsWith` would work, though, since it would also match on application/json+something.  I´d rather `StringUtils.substringBefore(contentType, ";")` and then make an equals-comparison

wilkinsona:
@openwms Thanks for the PR. I share @AGoliath's concern about the use of `startsWith`. Given the existing dependency on Spring Framework, I think the use of `MediaType.parseMediaType` is the safest option.

spring-restdocs/src/main/java/org/springframework/restdocs/hypermedia/LinkExtractors.java
```diff
@@ -17,18 +17,13 @@
 package org.springframework.restdocs.hypermedia;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
wilkinsona: Please do not use wildcard imports

spring-restdocs/src/test/java/org/springframework/restdocs/hypermedia/LinkExtractorsTests.java
```diff
@@ -16,27 +16,19 @@
 
 package org.springframework.restdocs.hypermedia;
 
-import static org.junit.Assert.assertEquals;
+import static org.hamcrest.core.IsNull.*;
```
wilkinsona: Please do not use wildcard imports

spring-restdocs/src/main/java/org/springframework/restdocs/hypermedia/LinkExtractors.java
```diff
@@ -70,10 +65,13 @@ public static LinkExtractor atomLinks() {
...
 	public static LinkExtractor extractorForContentType(String contentType) {
-		if (MediaType.APPLICATION_JSON_VALUE.equals(contentType)) {
+		if (null == contentType) {
+			return null;
+		}
+		if (contentType.startsWith(MediaType.APPLICATION_JSON_VALUE)) {
 			return atomLinks();
 		}
-		else if ("application/hal+json".equals(contentType)) {
+		else if (contentType.startsWith("application/hal+json")) {
```
wilkinsona: As mentioned in the main comments on the PR, I think this should use `MediaType.parseMediaType(contentType)`

openwms:
Perfect @wilkinsona . Fully agree with @AGoliath about the strictness. Sorry for the wildcard imports, thats a default behaviour of IntelliJ

wilkinsona:
@openwms Would you like to update the PR when you have a moment?

openwms:
Hey @wilkinsona please have a look at the updated 2 classes

wilkinsona:
Thank you. I've merged the changes.

PR #59, opened 2015-04-20
Fix broken include in samples

orfjackal:


wilkinsona:
Well spotted. Thank you.

PR #65, opened 2015-04-30
Create a more generic pattern-replacing post-processor

dewet22:
...and make the link masker extend it. Like I showed in the unit test, our use-case is actually to do some anonymising of data along the way.


wilkinsona:
Nice! Thanks very much.

PR #81, opened 2015-07-02
Configure Gradle to add generated documentation to resources not the jar directly

Endron:
While playing around with Spring Rest Docs and Spring Boot I run into a problem: When I run the application using

    ./gradlew bootRun

the generated documentation is not accessible on the embedded Tomcat server.
This is because the ``bootRun`` task from the Spring Boot Gradle Plugin does not depend on ``jar`` so copying the documentation using the jar task can not work.

This can be solved by copying the documentation earlier in the ``processResources`` task.


Endron:
Just realized this does in fact not work.
I need the tests to be completed to have the snippets. And I can not run tests without processing the resources.

Issue #119, opened 2015-09-09
Configure the build to use Checkstyle


wilkinsona:
The build's configured to use Sonar but this only catches problems after the fact (and reports a number of false positive). Adding check style to the build will allow problems to be caught before they're pushed.

PR #125, opened 2015-09-13
Configure the build to use Checkstyle

pcvolkmer:
This commit adds checkstyle plugin and provides a simple config file.

The configuration file is placed in projects root folder and should be
enhanced to contain all needed checks.
Some failing checks are disabled within this config file.

Fixes gh-119


wilkinsona:
Thanks

PR #126, opened 2015-09-14
Add detection charset from Content-type header in the HttpResponse

fossamagna:
When charset of text in content is not platform default charset, Content of HttpResponseSnippet is garbled.
If set charset with `Content-Type` header, I think HttpResponseSnippet should use charset attribute value to convert string from HTML body content.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/http/HttpRequestSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/http/HttpRequestSnippet.java
@@ -16,9 +16,9 @@
 
 package org.springframework.restdocs.http;
 
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


wilkinsona:
Thank you. I think this might be better handled by having a `getContentAsString` method on `OperationRequest` and `OperationResponse`.  That way, the conversion can be performed in a central place for the benefit of any snippets that want to use the content as a string. Would you be interested in updating this PR to do so?

fossamagna:
@wilkinsona :ok: I agree with you. I will update PR to do so.

fossamagna:
@wilkinsona I updated PR to have a `getContentAsString` method on `OperationRequest` and `OperationResponse`.
Please check PR.

wilkinsona:
Thanks again. Can you please [sign the CLA](https://github.com/spring-projects/spring-restdocs/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement)?

fossamagna:
@wilkinsona I have signed the CLA.

wilkinsona:
Thanks once again for the PR. I've merged this into master along with some additional changes to honour the charset wherever the contents of a request or response are worked with as a String.

PR #128, opened 2015-09-17
add sourceDir to asciidoc task inputs

otrosien:
I haven't signed the spring CLA yet, but I expect this to be just a "trivial patch".
asciidoctor does not register the sourceDir as task input, so the up-to-date check, or gradle continuous mode (-t) does not pick up changes in this directory.

### DIFF
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -16,4 +16,5 @@ asciidoctor {
 	attributes	'revnumber': project.version,
 				'branch-or-tag': project.version.endsWith('SNAPSHOT') ? 'master': "v${project.version}"
 	inputs.files(sourceSets.test.java)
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR. Testing with Gradle 2.6, continuous mode works fine for me in both the main docs and in the sample. Can you please describe what you were doing that you expected to trigger a build but it didn't?

otrosien:
Hi Andy, 

sorry, my mistake - I was running continuous mode in my project to and I didn't get the rebuild triggered. but now it seems to work without the need to change the gradle build. I also checked the asciidoctor plugin, they do use the InputDirectory annotation....

Anyway - keep up the good work :)

PR #144, opened 2015-10-03
Corrected asciidoc gradle task properties

sudhcha:
The asciidoctor gradle plugin didn't have an inputs.dir property. As a result, the gradle task was failing since it was trying on the default src/asciidoc directory. The correct property to override is sourceDir. The file is executing fine after this change.

### DIFF
--- a/docs/src/docs/asciidoc/getting-started.adoc
+++ b/docs/src/docs/asciidoc/getting-started.adoc
@@ -59,7 +59,7 @@ described below.
 
 	asciidoctor { <5>
 		attributes 'snippets': snippetsDir <6>
-		inputs.dir snippetsDir <7>
+		sourceDir snippetsDir <7>
 		dependsOn test <8>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, but the change you've made is incorrect.

The `sourceDir` is the location from which your main Asciidoc file(s) should be read. Your change means it will only find the individual snippets. Also, all tasks have `inputs.dir`, so whatever your problem was, that wasn't the cause.

Have you perhaps put your main Asciidoc file in the wrong place? When using Gradle they should go in `src/docs/asciidoc` by default.

sudhcha:
Thanks a lot. I misunderstood that even the main asciidoc would be created as part of the task (blame it on haste to implement something before reading the docs completely). Once I  corrected that, everything worked like a charm. 
Thanks for the clarification and this helpful project.

PR #157, opened 2015-10-23
Improvements to rest-assured module

yleguern:
Add path parameters & query parameters documentation capabilities.


wilkinsona:
Thanks for the PR. Unfortunately, I'm not keen on the approach you've taken for documenting path parameters, specifically the need to duplicate the path.

@johanhaleby Is there a way to "capture" the path in its original form, i.e. before any parameters have been replaced, so that they can be documented? The approach I took with MockMvc was to provide [a custom request builder](https://github.com/spring-projects/spring-restdocs/blob/master/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/RestDocumentationRequestBuilders.java). I don't think that approach will work so well with REST Assured due to the chaining of method calls.

johanhaleby:
@wilkinsona Hmm as far as I can remember you cannot do this in REST Assured but I'm willing to add it if required. Is all you need the path before applying the path parameters (for example /{x}/{y})?

yleguern:
I agree the approach is not as elegant as the one you used for mockmvc. Unfortunately I did not find a similar way for rest-assured : the only solution I found was to pass the path as the parameter. If you think of a better solution I'd be glad to implement it. 

Best regards.

johanhaleby:
I'll be happy to make changes in REST Assured to make integration with rest docs smoother.

yleguern:
That'd be great. 
Thanks Johan.

johanhaleby:
Just tell me what to do :)

wilkinsona:
@johanhaleby Could it be made available via the `FilterContext`?

johanhaleby:
@wilkinsona Sure. But should it include only the path or path + query? (see https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)

wilkinsona:
@johanhaleby Thanks. Just the path is sufficent for our purposes. The request spec's query parameters are already sufficient for dealing with the query string

johanhaleby:
May I ask what you think is a good name for the method in FilterContext? For example `getOriginalRequestPath` or  `getRequestPathWithPathParamNames`?

wilkinsona:
Looking at `FilterableRequestSpecification`, I'd be tempted to take inspiration from `getBasePath` and call it `getOriginalBasePath`. I'd also be tempted to add the method to `FilterableRequestSpecification` rather than on `FilterContext` as I originally suggested.

johanhaleby:
Thanks for your comments

johanhaleby:
I've added `getOriginalRequestPath` to `FilterContext` for now and published a new snapshot (2.6.1-SNAPSHOT). It requires the following repository if anyone want to try it out:

```xml
<repositories>
        <repository>
            <id>sonatype</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
            <snapshots />
        </repository>
</repositories>
```

@wilkinsona Have I understand it correctly that you're working with better support for third-party libraries such as RestAssured in the next version of spring rest docs or is it already implemented?! (I haven't tried it myself). If so, should I try to make a proper release of REST Assured with this fix asap?

wilkinsona:
This is great. Thanks, @johanhaleby. The REST Assured support is till a work in progress that'll hopefully be part of 1.1.0. There's not rush for a REST Assured release. We can build against snapshots in the meantime.

johanhaleby:
@wilkinsona Thanks! I'll probably release a new version in 1-2 weeks. Just tell me if there's anything else I can do to ease integration with rest docs.

johanhaleby:
@wilkinsona I've now released REST Assured 2.7.0 that contains the `getOriginalRequestPath` method in `FilterContext` that you should be able to use.

wilkinsona:
@johanhaleby Thanks very much

wilkinsona:
@yleguern Thanks again for the pull request. I didn't use it in the end for the reasons we already discussed above. However, it was still very useful as it led to the improvement to REST Assured that @johanhaleby kindly made so that the path parameters could be documented.

Issue #164, opened 2015-11-14
Improve support for documenting payloads with multiple sets of links


wilkinsona:
Inspired by [this question] on Stack Overflow from @matsev.

I had always assumed that links in embedded resources would be documented as part of the embedded resource's documentation, rather than in the embedding resource's documentation. That may not be an assumption that suits everyone. If you don't want to take this approach then the options currently available are:

1. Use a custom link extractor and produce a single snippet but this means that you can't assert that the links were in the right place in the payload
2. Use a custom snippet, probably a `LinksSnippet` subclass, and produce two separate snippets with different names.

matsev:
I have given this problem some more thought and there are some things that need special attention. Consider the following payload:

```json
{
    "foo": {
        "links": {
            "alpha": "http://alpha.example.com/one",
            "bravo": "http://bravo.example.com/one"
        }
    },
   "bar": {
        "links": {
            "alpha": "http://alpha.example.com/two",
            "bravo": ["http://bravo.example.com/one", "http://bravo.example.com/two"],
            "charlie": "http://charlie.example.com",
        }
    }
}
```
- Rel:s have the same semantic meaning if they have the same name
- The `links` part of the sub-documents have the same structure
- The `foo` and `bar` sub-documents differ
- The url of the `alpha` rel is different in the `foo` and `bar` sub-documents.
- The number of urls are different for the different `bravo` rel:s.

From a client perspective, this should not be a problem. However, from REST Docs' perspective the rel should only be documented once.

One approach could be to separate the description of rel:s from the the definition of links, e.g.
```java
Map<String, String> rels = rels(
    rel("alpha").description("Alpha description");
    rel("beta").description("Beta description");
    rel("charlie").description("Charlie description");
);

links(rels, 
    customLinks("$.foo.links"),
        linkWithRel("alpha"),
        linkWithRel("beta")),
    customLinks("$.bar.links"),
        linkWithRel("alpha"),
        linkWithRel("beta")
        linkWithRel("charlie")),
);
```

PR #165, opened 2015-11-14
Implemented LinkExecutor with json path selector

matsev:
Based on my question at [Stack Overflow](http://stackoverflow.com/questions/33680699/spring-restdocs-document-links-in-child-documents) and the feedback that I received, I thought I’d better share the solution I implemented in case you are interested. 

I noticed that the format of the links in our JSON response have a lot in common with the links in the JSON HAL format. Consequently, the `HalLinkExtractor` could be implemented as a specialization of the more flexibel format that I needed. Moreover, all tests in the `LinkExtractorsPayloadTests` could be applied to the `DynamicJsonPathLinkExtractor` without changes.

The current implementation of the `DynamicJsonPathLinkExtractor` uses a recursive algorithm to traverse the JSON document in order to find the links sub-document. There are both advantages and disadvantages with this approach:

Pros:
- No additional dependency required
- Has the same API as the existing HalLinkExtractor` (except for the constructor) and share a lot of code

Cons:
- Different syntax compared to [JsonPath](https://github.com/jayway/JsonPath) that is used in the MockMvc framework, e.g. does not use `$.` to identify the document root
- Does not support arrays, e.g. `$.alpha[0].links`
- Incapable of verifying in which sub-document a link is located, c.f. #164

I have also created a similar implementation using the `JsonPath` library. It is likely that many users of the Spring REST Docs already using `JsonPath` for JSON payload validation, so the addition of the library may not be a big deal.


matsev:
I have signed the CLA and contributed to Spring projects before, e.g. [DiskSpaceHealthIndicator](http://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/actuate/health/DiskSpaceHealthIndicator.html).

matsev:
I just pushed an implementation using the JsonPath library so that you can compare what the different implementations look like.

wilkinsona:
Thanks for the PR, but I haven't seen anyone else with this requirement so I'd prefer not to add support for it.

Issue #166, opened 2015-11-16
Add support for documenting attributes in XML payloads


cschaetzlein:
Given following xml
```xml
<persons>
  <person id="1">
    <firstname>John</firstname>
    <lastname>Doo</lastname>
  </person>
</persons>
```
trying to document the payload with

```java
fieldWithPath("/persons/person/@id").type("type").description("the persons id")
```

will lead to a NullPointerException at line 118

`node.getParentNode()` 
in `org.springframework.restdocs.payload.XmlContentHandler.getUndocumentedContent(List<FieldDescriptor> fieldDescriptors)`

since an attribute node has no associated parent node.

wilkinsona:
The current support for documenting XML payloads is only aimed at documenting elements rather than an individual element's attributes

PR #167, opened 2015-11-16
NPE when documenting xml attribute of xml payload

cschaetzlein:
Documenting a xml attribute in the xml payload leads to a
NullPointerException since no parent nodes exists for a xml attribute.
So check if node is an attribute and remove the attribute if so.

Fixes [gh-166](https://github.com/spring-projects/spring-restdocs/issues/166)


wilkinsona:
Thanks very much for the PR. If that's all it takes to fix it, we should be able to get this into 1.0.1. Have you [signed the CLA](https://github.com/spring-projects/spring-restdocs/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement)?

cschaetzlein:
Oh, sorry i missed that, signed the CLA a minute ago...

It is all to make documenting xml attributes available. Though detecting undocumented attributes is not part of that. Don't know if this is desirable...

wilkinsona:
I think that's fine, particularly for 1.0.1 as it would be a breaking change to require individual attributes to be documented.

PR #181, opened 2015-12-17
Fix asciidoctor-maven-plugin's phase

making:
The document (6) explains `prepare-package` should be used, but the example is `package`.

Actually, I had a trouble copying and pasting according to this doc.

### DIFF
--- a/docs/src/docs/asciidoc/getting-started.adoc
+++ b/docs/src/docs/asciidoc/getting-started.adoc
@@ -134,7 +134,7 @@ described below.
 				<executions>
 					<execution>
 						<id>generate-docs</id>
-						<phase>package</phase> <6>
+						<phase>prepare-package</phase> <6>
 						<goals>
 							<goal>process-asciidoc</goal>
 						</goals>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Sorry that you found this confusing. The need to use `prepare-package` if you want to include the documentation in the jar file is mentioned in the call out referenced by (6):

> (6) If you want to package the documentation in your project’s jar you should use the `prepare-package` phase.

It's also mentioned again in the [Packaging the documentation](http://docs.spring.io/spring-restdocs/docs/current/reference/html5/#getting-started-build-configuration-maven-packaging) section:

> First, configure the Asciidoctor plugin so that it runs in the prepare-package phase, as described above

If you have suggestions for how to make this clearer, I'd definitely like to hear them. However, I can't merge this change in its current form it'll mean that the sections mentioned above (among others) no longer make sense.

making:
Thank you for your comment.
I found I had misunderstood.

Updated PR to explain the phase explicitly.
How do you think?

wilkinsona:
Having thought about this some more, I think it might be best to update the documentation to always use `prepare-package` for the phase. It'll be a lot simpler as it'll mean one set of instructions that'll work for those who want to embed the documentation in the jar and those that don't.

wilkinsona:
Thanks for the PR. As discussed above, I've decided to address the problem by recommending that the Asciidoctor plugin is always bound to the `prepare-package` phase. See #218 for that change.

PR #186, opened 2016-01-22
This code compares String for reference equality using the ==.

mnhock:
The CLA has been signed.


wilkinsona:
Oops. Thanks very much. Did you use a tool to find this? Checkstyle's integrated into the build and the CI jobs also run Sonar. Neither of them caught it.

mnhock:
Sonar should have the bug pattern but i don`t know if its activated on default. I found this with findbugs.

wilkinsona:
> Sonar should have the bug pattern but i don`t know if its activated on default.

Thanks. It's activated in the profile I'm using. Now I'm worried what else I have a false sense of security about…

Issue #191, opened 2016-02-03
Allow HeaderRemovingOperatingProcessor to remove all headers


rhuss:
Sometimes headers are not really needed for the examples so it would be nice to have an option to remove them all without having to adapt the template.

wilkinsona:
How about an option that takes one or more regular expressions and removes every header that matches at least one of them? Then, to remove all headers, you could do something like this:

```java
Preprocessors.removeMatchingHeaders(".+")
```

rhuss:
would be perfectly fine. I can prepare a PR for this, it's an easy one ;-)

wilkinsona:
A PR would be much appreciated. Thanks.

PR #195, opened 2016-02-04
Add support for pattern when removing headers

rhuss:
`String` and `java.util.regex.Pattern` can both now be used as arguments to `removeHeaders` preprocessor method. When a Pattern is used then all matching headers are removed.

Fix gh-191

The CLA has been submitted.

### DIFF
--- a/docs/src/docs/asciidoc/customizing-requests-and-responses.adoc
+++ b/docs/src/docs/asciidoc/customizing-requests-and-responses.adoc
@@ -105,6 +105,8 @@ different replacement can also be specified if you wish.
 `removeHeaders` on `Preprocessors` removes any occurrences of the named headers
 from the request or response.
 
+`removeMatchingHeaders` on `Preprocessors` applies the given patterns on every header and
+removes them when matching.
 
 
 [[customizing-requests-and-responses-preprocessors-replace-patterns]]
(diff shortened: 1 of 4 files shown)


rhuss:
Alternatively an extra method `removeMatchingHeaders` could be added. 'guess its more or less a matter of taste (loose typing vs. API inflation)

wilkinsona:
Thanks for the PR.

I'm not keen on `removeHeaders` taking an `Object`. I do like that it lets you mix patterns and strings, but, IMO, that's outweighed by the loss of type safety. I'd prefer to either have `removeMatchingHeaders(String pattern)` or overload `removeHeaders` with `removeHeaders(Pattern pattern)`. I'm leaning towards the former as the user doesn't have to call `Pattern.compile` and the method name makes it clear that there's some matching involved.

rhuss:
Ok, updated the PR accordingly.

wilkinsona:
@rhuss Thanks very much for the PR. I've merged it into master. I made a few changes along the way. Please take a look at https://github.com/spring-projects/spring-restdocs/commit/1131c9fa6ee7137f6b85dbbf183b39eb3321dec8 if you're interested.

rhuss:
Thanks for picking it up and also for tuning the code. Looks indeed nicer that way.

PR #207, opened 2016-03-02
Add Httpie support in addition to curl

rocketraman:
Add support for httpie (http://httpie.org/) in addition to curl.

Httpie is "a CLI, cURL-like tool for humans" with features like automatic formatting and colorization of responses, and a clean command-line syntax that doesn't require things like escaping special characters in form fields. It works well for documenting an API.

I shared as much code from the curl implementation as possible by moving everything into a "cli" package, and creating an abstract base class from which both the curl and httpie implementation inherits.

I have signed the CLA (confirmation 165420160302074118).


spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/AbstractCliSnippet.java
```diff
@@ -0,0 +1,180 @@
...
+import org.springframework.restdocs.operation.OperationRequest;
+import org.springframework.restdocs.operation.Parameters;
+import org.springframework.restdocs.snippet.Snippet;
+import org.springframework.restdocs.snippet.TemplatedSnippet;
+import org.springframework.util.Base64Utils;
+
+/**
+ * An abstract {@link Snippet} that for CLI requests.
+ *
+ * @author Raman Gupta
```
wilkinsona: All of the authors of the original `CurlRequestSnippet` should also be credited here

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/AbstractCliSnippet.java
```diff
@@ -0,0 +1,180 @@
...
+		headerFilters.add(new BasicAuthHeaderFilter());
+		HEADER_FILTERS = Collections.unmodifiableSet(headerFilters);
+	}
+
+	/**
+	 * Create a new abstract cli snippet with the given name and attributes.
+	 * @param snippetName The snippet name.
+	 * @param attributes The snippet attributes.
+	 */
+	protected AbstractCliSnippet(final String snippetName, final Map<String, Object> attributes) {
```
wilkinsona: `snippetName` and `attributes` do not need to be `final`

wilkinsona:
@rocketraman Thanks for the PR. This looks excellent. I've made a few comments of the diff/commits for your consideration.

rocketraman:
> @rocketraman Thanks for the PR. This looks excellent. I've made a few comments on the diff/commits for your consideration.

Thanks for your review. I've made a bunch of the changes you requested, and commented on the others.

wilkinsona:
Unfortunately, I can't see your comments. I wonder if that's my fault for making some comments on your commits, rather than on the pull request diff? They may have got lost when you pushed up your latest changes. If you can remember what you said and where, would you mind posting the comments on the diff please?

rocketraman:
> I wonder if that's my fault for making some comments on your commits, rather than on the pull request diff

Yeah, when I force pushed the updates those comments were lost along with the commits. I pushed branch "httpie-old" from my reflog and thankfully it looks like GitHub kept the comments against the old hashes... see them here:

https://github.com/rocketraman/spring-restdocs/commits/httpie-old

wilkinsona:
Reflog saves the day once again. Thank you. I've added a few responses. I think we're almost there now.

rocketraman:
Round 2: everything resolved except for the filename issue with the process substitution. See my reply: https://github.com/rocketraman/spring-restdocs/commit/90b69f91fdbc258c05f8d51fca78be740f4bd406#commitcomment-16536698

wilkinsona:
Thanks again for the PR. I've merged it into master along with [a little bit of polish](https://github.com/spring-projects/spring-restdocs/commit/37e16bc308b08d465b0a3aee3adbeeaa70877b86).

rocketraman:
> Thanks again for the PR. I've merged it into master along with a little bit of polish.

Great!

PR #215, opened 2016-03-25
Restored the mention of Spring Framework 4.1 as a requirement

christopheml:
The mention of Spring Framework 4.1 as a requirement for Spring REST Docs was previously added in the readme (ea77bd5) has been removed when most of the content of the readme was moved to the documentation (8a0b075).

Most articles and documents about Spring REST Docs don't mention this, which means that people usually get stuck when tying to configure ```MockMvc```in their tests. Having this information in the readme is a nice time saver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ To learn more about Spring REST Docs, please consult the [reference documentatio
 
 ## Building from source
 
-Spring REST Docs requires Java 7 or later and is built using [Gradle][10]:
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, but I don't want to duplicate information from the docs in the README. The details in the README are intended to describe what's required to build the project, while the reference docs describe the requirements to use REST Docs. I've opened #217 to clarify things, including linking to the reference docs from the README.

PR #256, opened 2016-06-07
Re-order table of samples to make it easier for new users

dsyer:
If I'm just writing a simple service with Spring, I'm unlikely
to want to know about restassured. It's great that I have the choice
and I will look at it later, but I'd like to see the more Spring
oriented sampled at the top of the table. And the one that uses Maven.

### DIFF
--- a/docs/src/docs/asciidoc/getting-started.adoc
+++ b/docs/src/docs/asciidoc/getting-started.adoc
@@ -14,9 +14,15 @@ If you want to jump straight in, a number of sample applications are available:
 |===
 | Sample | Build system | Description
 
-| {samples}/rest-assured[REST Assured]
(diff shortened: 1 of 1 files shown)


jlstrater:
I disagree.  I like to see the table in the current alphabetical ordering.

wilkinsona:
I don't like the proposed ordering either and prefer alphabetical (although Grails is in the wrong place at the moment).

Perhaps it would make sense to split things up into three tables?

1. MockMvc (containing Spring Data REST and Spring HATEOAS)
2. REST Assured (containing Grails and REST Assured)
3. Advanced (containing Slate and TestNG)

wilkinsona:
Closing in favour of #266

Issue #260, opened 2016-06-14
Line Breaks in cURL and HTTPie Snippets


psamsotha:
It's pretty common to see cURL commands written with line breaks. For example

``` bash
$ curl 'http://localhost/foo' -i -X POST \
   -H 'X-Header-One: ONE' \
   -H 'X-Header-Two: TWO' \
   -d 'a=aplha&b=bravo'
```

It would be nice to have support for this, especially for longer snippets.

wilkinsona:
Nice idea. Thanks for the suggestion.

PR #261, opened 2016-06-14
Add Support for cURL Snippets with Line Breaks

psamsotha:
To add support for cURL snippets with line breaks, a few classes were added:
- `CurlPart` - represents a part of a cURL command
- `CurlLineGroup` - a group of cURL curl parts that make up a line
- `CurlLineBreakStrategy` - a strategy for how lines should
  be broken up
- `CurlLineBreakStrategies` - factory methods for strategies

To support configurability, the `CurlRequestSnippet` has been almost completely redone. Methods that originally wrote out a part of a command, have been abstracted into a into a class, for which instances can be mapped to the according `CurlPart`. The original code to write snippets was

``` java
private String getOptions(Operation operation) {
    StringWriter command = new StringWriter();
    PrintWriter printer = new PrintWriter(command);
    writeIncludeHeadersInOutputOption(printer);
    CliOperationRequest request = new CliOperationRequest(operation.getRequest());
    writeUserOptionIfNecessary(request, printer);
    writeHttpMethodIfNecessary(request, printer);
    writeHeaders(request, printer);
    writePartsIfNecessary(request, printer);
    writeContent(request, printer);

    return command.toString();
}
```

Each one of those `writeXxx` methods have been abstracted to a `WriteAction` class, that can be mapped to a `CurlPart`. How the cURL parts are grouped are through `CurlLineGroup`s. For example, if we have the following strategy implementation

``` java
CurlLineGroup optionsAndMethod = new CurlLineGroup(CurlPart.SHOW_HEADER_OPTION,
        CurlPart.USER_OPTION, CurlPart.HTTP_METHOD,
        CurlPart.MULTIPARTS);
CurlLineGroup headers = new CurlLineGroup(CurlPart.HEADERS);
CurlLineGroup content = new CurlLineGroup(CurlPart.CONTENT);
return Arrays.asList(optionsAndMethod, headers, content);
```

we have three line groups, each of which will be in a separate line. The order of the parts can also be switched around. Also, as headers and parts have multiple parts, we can also separate those individual parts on separate lines, by returning true or false on the following methods of the strategy interface

``` java
@Override
public boolean splitHeaders() {
    return true;
}

@Override
public boolean splitMultiParts() {
    return false;
}
```

So with the following complete implementation

``` java
private static final class HeadersAndContentCurlLineBreakStrategy
        implements CurlLineBreakStrategy {

    @Override
    public List<CurlLineGroup> getLinesGroups() {
        CurlLineGroup optionsAndMethod = new CurlLineGroup(CurlPart.SHOW_HEADER_OPTION,
                CurlPart.USER_OPTION, CurlPart.HTTP_METHOD,
                CurlPart.MULTIPARTS);
        CurlLineGroup headers = new CurlLineGroup(CurlPart.HEADERS);
        CurlLineGroup content = new CurlLineGroup(CurlPart.CONTENT);
        return Arrays.asList(optionsAndMethod, headers, content);
    }

    @Override
    public boolean splitHeaders() {
        return true;
    }

    @Override
    public boolean splitMultiParts() {
        return false;
    }
}
```

we would get the following output

``` bash
$ curl 'http://localhost/foo' -i -X POST \
   -H 'X-Header-One: ONE' \
   -H 'X-Header-Two: TWO' \
   -H 'X-Header-Three: THREE' \
   -d 'Some Content'
```

With this infrastructure in place, it is easy for the user to implement their own strategy, if none of the ones available suit their requirements.

The default strategy is to use no line breaks, so specifying none will be business as usual. There are a few other strategy implementations. You can see them in the `CurlLineBreakStrategies` class, for which there are the following static factory methods

``` java
none()
headersOnly()
partsOnly()
headersAndParts()
contentOnly()
headersAndContent()
```

See the javadoc for example output for each of those strategies.

Strategies can be plugged in simply by specifying an operation attribute key `CurlLineBreakStrategy.class.getName()`, with the value being an implementation of the `CurlLineBreakStrategy` interface. You can use one of the above factory methods, or just provide your own implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build
 target
 .idea
 *.iml
+/.nb-gradle/
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlLineBreakStrategies.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlLineBreakStrategies.java
@@ -137,7 +137,7 @@ public final class CurlLineBreakStrategies {
 			implements CurlLineBreakStrategy {
 
 		@Override
-		public List<CurlLineGroup> getLinesGroups() {
+		public List<CurlLineGroup> getLineGroups() {
 			return Arrays.asList(new CurlLineGroup(CurlPart.SHOW_HEADER_OPTION,
 					CurlPart.USER_OPTION, CurlPart.HTTP_METHOD,
 					CurlPart.HEADERS, CurlPart.MULTIPARTS, CurlPart.CONTENT));
@@ -158,7 +158,7 @@ public final class CurlLineBreakStrategies {
 			implements CurlLineBreakStrategy {
 
 		@Override
-		public List<CurlLineGroup> getLinesGroups() {
+		public List<CurlLineGroup> getLineGroups() {
 			CurlLineGroup allButHeaders = new CurlLineGroup(CurlPart.SHOW_HEADER_OPTION,
 					CurlPart.USER_OPTION, CurlPart.HTTP_METHOD,
 					CurlPart.MULTIPARTS, CurlPart.CONTENT);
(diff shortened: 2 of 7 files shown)


wilkinsona:
As I said on #260, I like the idea, however I have a feeling that this implementation is heavier than it needs to be. It is only a feeling at this stage. Sorry for not having anything more concrete at the moment, but I didn't want to appear to be ignoring this PR and I haven't had time to investigate whether or not my feeling is right or not.

psamsotha:
@wilkinsona No worries. I know how it is with non trivial PRs. I kinda had the same feeling, about it being bloated. Maybe I was trying to make it _too_ configurable.

PR #266, opened 2016-06-21
rearrange samples table into three tables by type

jlstrater:
closes #256


wilkinsona:
Thanks very much, @jlstrater

PR #268, opened 2016-06-21
Add README to grails example closes #254

### LINKED ISSUE #254: Missing documentation/explanation of samples
I am trying out the `rest-notes-grails` sample project.

I am running
```
gradle build
```
Which generates
```
build/asciidoc/index.html
```

I assume that this file is the entry point for the documentation. However it contains a lof of "dead links" adoc references to files that does not exist. I don't know which of the below is most likely:

1. I am doing something wrong, in which case I don't know what.
1. The example is not yet mature enough to actually show the complete process to generate the missing files.
1. There is a "bug" in the sense that the example is believed to show a complete workflow, but it aparently fails... (Maybe someone failed to commit some files, what do I know).

In any of the above case, it would be nice to have an explanation in the README.md for that example.

jlstrater:
Does this resolve your issue with the Grails documentation @jarl-dk ? Or is there anything else you'd like me to add?

### DIFF
--- a/samples/rest-notes-grails/README.md
+++ b/samples/rest-notes-grails/README.md
@@ -1,2 +1,85 @@
 # grails-spring-restdocs-example
-Example of using Spring REST Docs with Grails
+
+## Overview
+
+This is a sample project using Grails 3, Spock, and Spring REST docs.  For more information about the Grails framework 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @jlstrater

Issue #267, opened 2016-06-21
Give a way to document Multipart content


mathieu-pousse:
I have an endpoint to upload a file and its associated metadata.

It is a multipart request with: 
 * "file": "application-octect-stream"
 * "metadata": application/json

I am able to document the parts but I can't find a way to document the content of the metadata part: 

```
.andDo(document("upload",
    getDocRequestPreprocess(simplifyFileUpload("file", "<<zip content>>\n\n")),
    getDocResponsePreprocess(),
        RequestDocumentation.requestParts(
            partWithName("file").description("the file to upload"),
            partWithName("metadata").description("json that describes the file to import"))
            // How to document the metadata part content ?
    ));
```

```
----
POST /v1/somewhere/
Content-Type: multipart/form-data; boundary=6o2knFse3p53ty9dmcQvWAIx1zInP11uCfbm
Authorization: Bearer [Encoded-Bearer]
Host: localhost

--6o2knFse3p53ty9dmcQvWAIx1zInP11uCfbm
Content-Disposition: form-data; name=file
Content-Type: application/octet-stream

<<zip content>>


--6o2knFse3p53ty9dmcQvWAIx1zInP11uCfbm
Content-Disposition: form-data; name=metadata
Content-Type: application/json

{"something":10,"somethingElse":"matter"}
--6o2knFse3p53ty9dmcQvWAIx1zInP11uCfbm--
----
```

I think we need a way to set the part on which applies the FieldDescriptor or a way to document more precisely the part content

mathieu-pousse:
I'll submit a PR as I finally managed to implement this one

wilkinsona:
> I think we need a way to set the part on which applies the FieldDescriptor or a way to document more precisely the part content

Agreed. https://github.com/spring-projects/spring-restdocs/issues/161 was a small step in that direction, but there's plenty of room to do more. The hardest part will probably be getting the API right. Suggestions for how you'd like it to look would be most welcome.

wilkinsona:
> I'll submit a PR as I finally managed to implement this one

Great stuff! Thanks

mathieu-pousse:
here it is https://github.com/spring-projects/spring-restdocs/pull/270

any comment will be appreciated

M.

wilkinsona:
Closing in favour of the PR (#270).

PR #270, opened 2016-06-23
extend the request fields to be able to document the request part's content

mathieu-pousse:
see #267

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/RequestPartFieldsSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/RequestPartFieldsSnippet.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2014-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.restdocs.payload;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.http.MediaType;
+import org.springframework.restdocs.operation.Operation;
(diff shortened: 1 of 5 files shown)


spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/RequestPartFieldsSnippet.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	@Override
+	protected byte[] getContent(Operation operation) throws IOException {
+		for (OperationRequestPart candidate : operation.getRequest().getParts()) {
+			if (candidate.getName().equals(this.partName)) {
+				return candidate.getContent();
+			}
+		}
+		return new byte[0];
```
wilkinsona: I wonder if this should fail fast here rather than failing later when a documented field in the part isn't found?

wilkinsona:
@mathieu-pousse This looks excellent. Thank you very much. I've made one comment on the commit.

If you haven't done so already, can you please [sign the CLA](https://github.com/spring-projects/spring-restdocs/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement)?

mathieu-pousse:
I see.

I did not implement anything here because I think users will firstly document the parts with org.springframework.restdocs.request.RequestDocumentation#partWithName and then document the content of the part. 

That implies org.springframework.restdocs.request.RequestPartsSnippet#verifyRequestPartDescriptors has already handled missing parts.

Let me know

wilkinsona:
That's certainly possible, but may not always be the case. I think it would be more in keeping with the rest of the project if it failed fast with an exception messaging indicating that the part doesn't exist.

mathieu-pousse:
yep, i am adding test cases and will add this exception.
Can I throw the same error as here
? org.springframework.restdocs.request.RequestPartsSnippetFailureTests#missingPart
?

On Fri, Jun 24, 2016 at 10:24 AM, Andy Wilkinson <notifications@github.com>
wrote:

> That's certainly possible, but may not always be the case. I think it
> would be more in keeping with the rest of the project if it failed fast
> with an exception messaging indicating that the part doesn't exist.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/spring-projects/spring-restdocs/pull/270#issuecomment-228286771>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AChQZwMP2MDvICraaDOLdmqKjMgANqthks5qO5RTgaJpZM4I88DY>
> .
>



-- 

*Mathieu POUSSE | **06.28.89.08.43** |* *@m_pousse*
*Zenika Ouest*
18, rue de la Monnaie
35000 RENNES

wilkinsona:
Yep, that'd be perfect. Thanks!

mathieu-pousse:
here we are

wilkinsona:
@mathieu-pousse Thanks again for the PR. The changes are now in master.

PR #280, opened 2016-07-21
Remove unnecessary Host header set in UriModifyingOperationPreprocessor

izeye:
The `Host` header set value will be replaced below without using it, so it looks unnecessary.


wilkinsona:
Thanks!

Issue #297, opened 2016-09-02
Provide a convention-based approach for determining the output directory


wilkinsona:
It'd be nice not to have to configure the snippet output directory, and certainly not in two places (build script and the tests) as is currently required. The former is used to tell Asciidoctor from where the snippets can be included. The latter is used to tell the tests where to write the snippets.

The output directory varies depending on which build system is in use. For Maven it's typically `target/generated-snippets`. For Gradle it's typically `build/generated-snippets`. We can determine which build system is in use by looking for a `pom.xml` in the current directory. If one is found then Maven is in use, otherwise assume that it's a Gradle build.

Making the tests use this convention-based output directory by default is straightforward. It's a simple change to `ManualRestDocumentation` and the wrappers that use it as a delegate to make the output directory optional. When one is not specified, the convention-based default should be used.

Getting the convention-based default into Asciidoctor isn't quite so straightforward. The current plan is to provide a custom Asciidoctor macro for snippet inclusion. This macro will include the logic for determining the output directory by convention. Rather than including a snippet like this:

```
include::{snippets}/error-example/http-response.adoc[]
```

It would be included like this instead:

```
snippet::error-example/http-response.adoc[]
```

The custom `snippet` macro will behave exactly like the `include` macro but will prefix the path with the automatically determined output directory.

meistermeier:
This is a really great idea for more convenience. Has there been any work on this issue? I would like to help and hack (especially but not only the asciidoctor part) ;)

wilkinsona:
Nothing of any real substance. A contribution would be most welcome.

meistermeier:
Here is my (not so small) change to get a default output directory handling. I did not extend or provide a new sample since I won't change too much before getting a review ;)
To use the new macro you have to include it in maven/gradle as a dependency:

    <plugin>
        <groupId>org.asciidoctor</groupId>
        ....
        <dependencies>
            <dependency>
	        <groupId>org.springframework.restdocs</groupId>
	        <artifactId>spring-restdocs-asciidoctor-extensions</artifactId>
	        <version>1.2.0.BUILD-SNAPSHOT</version>
            </dependency>
        </dependencies>
    ....
    </plugin>

or gradle

    dependencies {
        asciidoctor 'org.springframework.restdocs:spring-restdocs-asciidoctor-extensions:1.2.0.BUILD-SNAPSHOT'
    }

wilkinsona:
[...]

PR #310, opened 2016-10-09
Provide a default output directory for the snippets. Fixes gh-297.

meistermeier:
The path <projectDirectory>/<gradle/mavenDirectory>/generated-snippets will be
respected by a) a new block macro and b) the spring-restdocs-core class ManualRestDocumentation.
This behaviour can be overridden by a system property called snippetOutputDirectory.

a) The new block macro module (spring-restdocs-asciidoctor-extensions) is a dependency
for asciidoctors maven or gradle plugin. It provides the class RestDocsSnippetBlockMacro
that includes and let asciidoctor process the content from the default (or custom)
folder if the macro restdocs:: is used.

b) The ManualRestDocumentation class and the classes using it as a delegate were
modified. You cannot provide the documentation path as constructor parameter anymore.
Instead it uses the same mechanisms as a).

All tests (also the tests of the samples) were changed to the new signature.

There is no change in the documentation. This will happen if this commit / pull request
got reviewed.


pivotal-issuemaster:
@meistermeier Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-restdocs&pullRequestId=310)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-restdocs&pullRequestId=310) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@meistermeier Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-restdocs&pullRequestId=310)!

meistermeier:
take a look at my issue comment to see how-to-use https://github.com/spring-projects/spring-restdocs/issues/297#issuecomment-252513847

meistermeier:
After I realized how the imports are ordered in the project I thought it might be a benefit to have the same in the samples. Is there any convention to use a "fixed" ordering in the productive code but not in the samples?
It is ok for me to revert this since it is not the scope of the PR/issue and split it in two PRs :)

wilkinsona:
>  Is there any convention to use a "fixed" ordering in the productive code but not in the samples?

Unfortunately not. The samples are different for no better reason than they haven't been set up with the same code formatter settings.

> It is ok for me to revert this since it is not the scope of the PR/issue and split it in two PRs :)

Yes, of course. Thanks.

meistermeier:
Update done. I don't know if there is any automated github notification ;)

spring-restdocs-core/src/main/java/org/springframework/restdocs/JUnitRestDocumentation.java
```diff
@@ -34,12 +34,10 @@
 
 	/**
 	 * Creates a new {@code JUnitRestDocumentation} instance that will generate snippets
-	 * to the given {@code outputDirectory}.
-	 *
-	 * @param outputDirectory the output directory
+	 * to &lt;gradle/maven build path&gt;/generated-snippet if no runtime property ({@code snippetOutputDirectory}) is set.
 	 */
-	public JUnitRestDocumentation(String outputDirectory) {
```
wilkinsona: This constructor needs to be kept, both to allow the output directory to be overridden and to maintain backwards compatibility.

spring-restdocs-core/src/main/java/org/springframework/restdocs/ManualRestDocumentation.java
```diff
@@ -31,17 +33,31 @@
...
+		this(System.getProperty(OUTPUT_DIR_PROPERTY_KEY, getDefaultOutputDirectory()));
+	}
+
+	/**
+	 * Creates a new {@code ManualRestDocumentation} instance that will generate snippets
 	 * to the given {@code outputDirectory}.
 	 *
 	 * @param outputDirectory the output directory
 	 */
-	public ManualRestDocumentation(String outputDirectory) {
```
wilkinsona: This constructor needs to remain public, both to allow the output directory to be overridden and to maintain backwards compatibility.

spring-restdocs-core/src/main/java/org/springframework/restdocs/RestDocumentation.java
```diff
@@ -34,12 +34,10 @@
 
 	/**
 	 * Creates a new {@code RestDocumentation} instance that will generate snippets to the
-	 * given {@code outputDirectory}.
-	 *
-	 * @param outputDirectory the output directory
+	 * to &lt;gradle/maven build path&gt;/generated-snippet if no runtime property ({@code snippetOutputDirectory}) is set.
 	 */
-	public RestDocumentation(String outputDirectory) {
```
wilkinsona: While this class is deprecated (and might be removed before 1.2 is released), for the time being at least this constructor needs to be kept, both to allow the output directory to be overridden and to maintain backwards compatibility.

.gitignore
```diff
@@ -7,4 +7,5 @@ build
 !buildSrc/src/main/groovy/org/springframework/restdocs/build/
 target
 .idea
-*.iml
\ No newline at end of file
+*.iml
+.DS_Store
```
wilkinsona: This is OS-specific so it doesn't belong in the repository's `.gitignore`. It should go in your own [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead.

config/checkstyle/checkstyle-suppressions.xml
```diff
@@ -5,4 +5,6 @@
 <suppressions>
 	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocVariable" />
 	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocMethod" />
+	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocMethod" />
+	<suppress files="RestDocsSnippetBlockMacro\.java" checks="RedundantModifier" />
```
wilkinsona: Why is this suppression necessary?

config/checkstyle/checkstyle-suppressions.xml
```diff
@@ -5,4 +5,6 @@
 <suppressions>
 	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocVariable" />
 	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocMethod" />
+	<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="JavadocMethod" />
```
wilkinsona: This looks like a duplicate of line 7.

spring-restdocs-asciidoctor-extensions/build.gradle
```diff
@@ -0,0 +1,15 @@
+description = 'Spring REST Docs Asciidoctor Extensions'
+
+dependencies {
+	compile 'org.asciidoctor:asciidoctorj:1.5.2'
+	optional 'junit:junit'
+}
+
+task copyTestAsciidoctorFile(type: Copy) {
```
wilkinsona: If this copy is necessary, I'd rather do it as part of the test than in the Gradle build so that it happens when running things in an IDE too.

spring-restdocs-asciidoctor-extensions/src/main/java/org/springframework/restdocs/asciidoctor/extensions/RestDocsSnippetBlockMacroExtension.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author Gerrit Meier
+ */
+public class RestDocsSnippetBlockMacroExtension implements ExtensionRegistry {
+
+	/**
+	 * the name that identifies the block macro in the asciidoctor document
+	 * (e.g. <pre><b>restdocs</b>::file_to_include[]</pre>)
+	 */
+	private static final String BLOCK_NAME = "restdocs";
```
wilkinsona: `restdocs` is too generic as we may want to add more macros in the future. I think the macro should be called `snippet`.

spring-restdocs-asciidoctor-extensions/src/main/java/org/springframework/restdocs/asciidoctor/extensions/RestDocsSnippetBlockMacroExtension.java
```diff
@@ -0,0 +1,40 @@
...
+import org.asciidoctor.Asciidoctor;
+import org.asciidoctor.extension.spi.ExtensionRegistry;
+
+/**
+ * ExtensionRegistry for the {@link RestDocsSnippetBlockMacro} to get registered
+ * as an extension in all projects that include the asciidoctor extension module.
+ *
+ * @author Gerrit Meier
+ */
+public class RestDocsSnippetBlockMacroExtension implements ExtensionRegistry {
```
wilkinsona: I don't think this class should be named as if it's specific to a particular macro. IMO, `RestDocsExtensionRegistry` is a better name.

spring-restdocs-core/src/main/java/org/springframework/restdocs/ManualRestDocumentation.java
```diff
@@ -31,17 +33,31 @@
...
 	private final File outputDirectory;
 
 	private RestDocumentationContext context;
 
 	/**
 	 * Creates a new {@code ManualRestDocumentation} instance that will generate snippets
+	 * to &lt;gradle/maven build path&gt;/generated-snippet if no runtime property ({@code snippetOutputDirectory}) is set.
+	 */
+	public ManualRestDocumentation() {
+		this(System.getProperty(OUTPUT_DIR_PROPERTY_KEY, getDefaultOutputDirectory()));
```
wilkinsona: I don't like the use of a system property.

REST Docs used one in the past but it didn't work well when running individual tests in an IDE. Typically, the system property is only set in `build.gradle` or `pom.xml` so you ended up with different configuration.

For now at least, I'm happy for people who don't want to use the conventional locations to have to override the location using the constructor argument to `JUnitRestDocumentation` or `ManualRestDocumentation` and continuing to use `include::`.

config/checkstyle/checkstyle-suppressions.xml
meistermeier: Asciidoctor needs to instantiate the macro using reflection(?). If I reduce the visibility of the constructor I get `org.jruby.exceptions.RaiseException: (TypeError) asciidoctor: FAILED: <stdin>: Failed to parse source, no public constructors for Java::OrgSpringframeworkRestdocsAsciidoctorExtensions::RestDocsSnippetBlockMacro`. I don't see any possibility to use another registration method and also provide the macro's block name.

meistermeier:
I resolved most of the review points (see https://github.com/spring-projects/spring-restdocs/pull/310#discussion_r82660211). Now the changes should be backward compatible and I removed the system property as a configuration option.

wilkinsona:
Not at all. Thank you for the contribution.

There are just a couple of things missing now:

1. Updates to the samples to use the new macro
2. Updates to the docs

I'm happy to take care of those as part of merging this, but if you'd like to tackle them it'd be much appreciated.

meistermeier:
Started with the samples and found a bug in the plugin (or Asciidoctor(J)?) if I use the macro in two separated files (e.g. docs/docA.asc, docs/docB.asc) the second one will throw an exception on the first occurrence/execution of the macro. I will investigate the problem in the next days.
For documentation purpose an excerpt of the stacktrace:

    Caused by: org.apache.maven.plugin.PluginExecutionException: Execution generate-docs of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc failed: org.jruby.exceptions.RaiseException: (TypeError) can't dup NilClass
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:145)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
	... 27 more
    Caused by: org.asciidoctor.internal.AsciidoctorCoreException: org.jruby.exceptions.RaiseException: (TypeError) can't dup NilClass
	at org.asciidoctor.internal.JRubyAsciidoctor.renderFile(JRubyAsciidoctor.java:345)
	at org.asciidoctor.maven.AsciidoctorMojo.renderFile(AsciidoctorMojo.java:289)
	at org.asciidoctor.maven.AsciidoctorMojo.execute(AsciidoctorMojo.java:180)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
	... 28 more
    Caused by: org.jruby.exceptions.RaiseException: (TypeError) can't dup NilClass
	at org.jruby.RubyKernel.dup(org/jruby/RubyKernel.java:2029)
	at RUBY.(:1)
	at RUBY.(:1)
	at org.jruby.gen.InterfaceImpl175128506.convertFile(org/jruby/gen/InterfaceImpl175128506.gen:13)

wilkinsona:
Thanks for the PR, Gerrit. I've merged in https://github.com/spring-projects/spring-restdocs/commit/3ac4a1acada45cacc0a7d394edf719315f26b054 and made some tweaks in 319bd13.

meistermeier:
I also think that this is an cleaner solution than doing the conversion by hand that results in only HTML support. You could have still use PDF or something else but the old macro would have generated HTML content.
@roberpanzer and I chatted about this pull request and the possibilities in asciidoctor to use other backends with my block macro. There will only be a chance with asciidoctorj 1.6. to use other contexts than _pass_ to get asciidoctor to "compile" the file content without wrapping it in some kind of section.
There is still a problem with your changes and it is the same I mentioned above:
It will only work if you start maven or gradle in the project directory. It will search for the files relative to the execution directory / PWD.
One way to archive this is to use `document.getAttr("docfile")` or `document.getAttr("docdir")` (or maybe some other attribute you can find [here in the manual](http://asciidoctor.org/docs/user-manual/). Also thanks to Robert for giving me this hint.

wilkinsona:
I'm using `docdir` with Maven combined with the working directory to get a relative path from the `.adoc` file to the snippets. In my testing, this seems to work with multi-module builds but I may have missed something. I'm using `projectdir` with Gradle which makes things very straightforward. This all happens in https://github.com/spring-projects/spring-restdocs/blob/master/spring-restdocs-asciidoctor/src/main/java/org/springframework/restdocs/asciidoctor/SnippetsDirectoryResolver.java.

wilkinsona:
@meistermeier I've just realised that I missed your update above about having fixed the bug and the directory problem. Sorry. Out of interest, what was the cause of the bug?

meistermeier:
Since I had to convert the file manually I needed a new instance of asciidoctor. After a few days thinking it has to depend on the options object in ruby, because this was the only object that was "dupped" in the source files, I realized that there is a create method that accepts the class path as a parameter. With this set to the class path of the current class it was working again.

PR #311, opened 2016-10-10
Add extension points to TemplatedSnippet for output file locations

dsyer:
This is a useful base class for custom snippets but it only knows
how to put the output in one place (operation.name/snippetName).
This change extracts those two locations (base dir and file name)
into protected methods without changing the existing behaviour.


wilkinsona:
Without making some other changes, this isn't quite right. `TemplatedSnippet` deliberately doesn't think in terms of files. It just cares about a `Writer` then can write output to wherever you want. The change proposed here would make `TemplateSnippet` aware of file-based output.

I see the need for some more flexibility, though. At the moment the snippet name is used for two purposes:

1. To find the snippet's template
2. To resolve the `Writer`

Would separating these two so that the same template can be used to write multiple, differently named snippets do what you need?

dsyer:
Probably. It comes to the same thing in the end. We need the template location and the output file not to share a name necessarily.

wilkinsona:
Ok. Thanks for the PR. I've opened #320 to implement the alternative approach.

Issue #325, opened 2016-10-27
Ignoring port value in the spring rest doc snippets


JananiSubbiah:
I am generating spring rest docs using mock mvc and I see that my doc always contains a port value (its either 8080 if I do not override it with a custom one or its the custom port value I provide). I am not sure if this is possible but I was looking for a way to omit the port from my request and response (in the snippets particularly). 
 For example my curl-request.adoc looks like this (I have set no port value)

```
[source,bash]
----
$ curl 'https://mylocalhost:8080/myapp' -i
----
```

and the @Before method in my junit looks like this:

```
this.mockMvc = MockMvcBuilders.webAppContextSetup(this.context).apply(documentationConfiguration(this.restDocumentation).uris().withScheme("https").withHost("mylocalhost")).build();
```

Is it possible for the curl-request.adoc to look as follows without the port?

```
[source,bash]
----
$ curl 'https://mylocalhost/myapp' -i
----
```

Thanks for any help in this regard!

wilkinsona:
Yes, just set it to the default for the scheme. That's 80 for http and 443 for https. This has been [asked before](http://stackoverflow.com/questions/39729544/how-to-remove-port-number-completely-from-example-curl) so the documentation probably needs to be improved.

JananiSubbiah:
I apologize for overlooking that question. Thanks a lot :)

wilkinsona:
No apology necessary! As I said, it's a good indicator that the docs could be better.

PR #328, opened 2016-11-02
Provide default port / scheme information in documentation

meistermeier:
Add a note about default ports 80 and 443 in combination with http or https scheme.
Fixes gh-325

PR #329, opened 2016-11-03
Provide default port / scheme information in documentation

meistermeier:
Add a note about default ports 80 and 443 in combination with http or https scheme.
Fixes gh-325

Same one as the https://github.com/spring-projects/spring-restdocs/pull/328 but now with `1.1.x` as base branch.


wilkinsona:
Thank you

PR #330, opened 2016-11-16
Add context path as modifiable URI part for RestAssured

wjans:
Also expose context path on the URI modifying preprocessor for RestAssured.

### DIFF
--- a/spring-restdocs-restassured/src/main/java/org/springframework/restdocs/restassured/operation/preprocess/UriModifyingOperationPreprocessor.java
+++ b/spring-restdocs-restassured/src/main/java/org/springframework/restdocs/restassured/operation/preprocess/UriModifyingOperationPreprocessor.java
@@ -69,6 +69,8 @@ public final class UriModifyingOperationPreprocessor implements OperationPreproc
 
 	private String port;
 
+	private String contextPath;
+
 	/**
 	 * Modifies the URI to use the given {@code scheme}. {@code null}, the default, will
 	 * leave the scheme unchanged.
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR. What's the problem that it is trying to address? I had imagined that a service could be tested locally using whatever context path is required and the host, scheme, and port can then be modified to reflect the real service.

wjans:
We are using Zuul as a gateway for our microservices. So we write our tests locally and we would like to modify the URI to the URI as it is exposed on the gateway, i.e. including the context path as it is configured in our Zuul routes.

wilkinsona:
Rather than modifying the URI to have a different context path, why not just run your micro service locally with that context path when you're generating its documentation?

wjans:
You mean by overriding it in the test configuration to 'fake' the path being used by the gateway?  I thought it would be nicer to have it explicitly defined.  It would also allow me to specify the zuul specific configuration all together in the same place (scheme/host/port/path).

wilkinsona:
> You mean by overriding it in the test configuration to 'fake' the path being used by the gateway?

Yes. For example, if it were a Spring Boot app, you'd just set `server.context-path=/foo` when running the tests that generate the documentation.

> It would also allow me to specify the zuul specific configuration all together in the same place (scheme/host/port/path).

I can see the benefit to that, but, assuming that you can "fake" the context path, I'm not yet convinced that the added complexity in REST Docs is worth it.

wjans:
Ok, I will rewrite my tests like you suggest, and maybe create some reusable factory methods reading the zuul configuration, which will also allow me to have it in one place.

PR #334, opened 2016-12-16
Clarify parameter {ClassName} - Simple class name

stonio:
Page http://docs.spring.io/spring-restdocs/docs/current/reference/html5/#documentating-your-api-parameterized-output-directories
Clarify that class name is not the default (fully-qualified) class name but the **simple** class name.


wilkinsona:
Thank you

Issue #304, opened 2016-09-13
Include cookies in the HTTP request snippet


wilkinsona:
When a request is created with cookies, the HTTP request snippet should include a `Cookie` header for each cookie.

Issue #303, opened 2016-09-13
Include cookies in the httpie request snippet


wilkinsona:
When a request is created with cookies, they should be include in the httpie request snippet as a `Cookie` header

PR #336, opened 2016-12-20
Include cookies in request snippets

### LINKED ISSUE #302: Include cookies in curl request snippet
When a request is created with cookies, they should be include in the curl request snippet using the `--cookie` option

tkopczynski:
This adds cookies to curl, HTTP and httpie request snippets.

In HTTP and httpie snippets it simply adds a `Cookie` header for each cookie. In the curl snippet it uses the `--cookie` option.

Fixes #302 #303 #304

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CliOperationRequest.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CliOperationRequest.java
@@ -25,6 +25,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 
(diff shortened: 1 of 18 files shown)


wilkinsona:
Nice! Thanks for the PR

tkopczynski:
@wilkinsona I'm glad you like it.
I'd like to continue this with a PR to support cookies in response snippets. It feels that I would need to use some of the code written already here, so what are your plans on merging this?

wilkinsona:
@tkopczynski Soon, I hope. I'm just catching up after the holidays. Sorry for the delay, but it will definitely be in 1.2.0.RC1.

Issue #305, opened 2016-09-13
Include cookies in the HTTP response snippet


wilkinsona:
When a response that contains cookies is set, each cookie should result in a `Set-Cookie` header appearing in the HTTP response snippet.

Issue #306, opened 2016-09-13
Provide support for documenting request and response cookies


wilkinsona:
There are a couple of ways in which this could be implemented:
- By adding `Cookie` and `Set-Cookie` headers for each cookie in the request and response respectively and using the existing headers snippet
- By adding specific request cookies and response cookies snippets and a richer cookie descriptor

I'm leaning towards the latter at the moment as it would allow the details of the cookie, even if it was just the name, to be verified.

PascaleBorscia:
Hello,

I prefer also the second implementation. A cookie is a special header and is used for special requests that require session affinity.
We use only the name in your application (identity). But I think it is interesting also to add  the path and the max-age as details.

Thank-you for your prompt response.

welcor:
I also think the second approach is most valuable. Having a "Cookie" header is no guarantee that there is a specific cookie value in it.

clydebarrow:
Although cookie data is now available in the response object, there is no support I can see for generating snippets for cookies - i.e. I would have expected a class like `ResponseCookiesSnippet` to be available. Is there any progress on this?

wilkinsona:
No progress has been made. I'm still leaning towards the latter approach, and that seems to be an opinion that's shared by you and those that have commented above. If you'd like to propose an implementation, I'd be more than happy to review a pull request.

wilkinsona:
Closing in favour of #592.

Issue #286, opened 2016-07-29
Query string is duplicated in generated snippets


jkubrynski:
After upgrading to 1.1.1.RELEASE query string is duplicated in generated snippets:

1.1.0.RELEASE `exams?count=10&page=1`

1.1.1.RELEASE `exams?count=10&page=1?count=10&page=1`

wilkinsona:
There are several tests around query strings that don't have any duplication so it's not immediately obvious what the cause may be. Can you please provide some more details:

- Which snippets are affected?
- Are you using MockMvc or RestAssured?
 - How are you creating the request?

A small sample that reproduces the problem would be ideal

jkubrynski:
Almost all snippets, including `curl-request.adoc` or `http-request.adoc`. I'm using MockMvc:
```
mockMvc.perform(
  get('/exams')
    .header("name", "value")
    .param('count', '10')
    .param('page', '1')
    .accept(MediaType.APPLICATION_JSON))
  .andDo(document(....)
```

wilkinsona:
Thanks, I can see where the problem is.

wilkinsona:
@jkubrynski This should be fixed in the latest 1.1.2 snapshots available from https://repo.spring.io/libs-snapshot. Early testing of a snapshot would be much appreciated.

jkubrynski:
Thanks @wilkinsona !

I confirm - it works correctly now :)

gdboling:
I'm running into this problem. Can someone tell me how to use the latest snapshot where this is fixed via gradle?

jkubrynski:
@gdboling 
```
dependencies {
    compile 'org.springframework.restdocs:spring-restdocs-mockmvc:1.1.2.BUILD-SNAPSHOT'
} repositories {
    maven {
        url 'https://repo.spring.io/libs-snapshot'
    }
}
```

gdboling:
Hm, I'm still getting duplicate request params:

```
[source,bash]
----
$ curl 'http://localhost:8080/api/notifications?type=A?type=A' -i -H 'X-Auth-Company: ABC' -H 'X-Auth-User: 1234' -H 'Content-Type: application/json;charset=UTF-8'
----
```
Here's my test
```
MockHttpServletRequestBuilder getRequest = get(URL);
        getRequest.header("X-Auth-Company", "ABC");
        getRequest.header("X-Auth-User", "1234");
        getRequest.param("type", "A");

[...]

PR #340, opened 2017-01-05
Include cookies in response snippets

tkopczynski:
This PR adds response cookies handling for Mock MVC and a few tests for handling `Set-Cookie` header for Rest Assured.

Closes gh-305

### DIFF
--- a/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcResponseConverter.java
+++ b/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcResponseConverter.java
@@ -16,6 +16,8 @@
 
 package org.springframework.restdocs.mockmvc;
 
+import javax.servlet.http.Cookie;
+
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.mock.web.MockHttpServletResponse;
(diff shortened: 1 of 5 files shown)


wilkinsona:
This one's proving tricker than I thought it would. I think I'd like the response cookies to be available from `OperationResponse` in the same way that they are now available from `OperationRequest`. It's not totally straightforward, though.

With the cookies available via a `getCookies()` method we'd have to figure out what to do with the `Set-Cookie` header. With REST Assured there'd be the same information in two places. With MockMvc, the cookies would just be available from `getCookies()` unless we explicitly set the header as well as is being proposed here. Having the same information in two places has caused [problems in the past](https://github.com/spring-projects/spring-restdocs/issues/286) so we need to tread carefully.

I'll give it some more thought.

tkopczynski:
I was thinking about this doubled information as well. I came to the conclusion that cookie information should be expressed as a header in OperationResponse (this is why I didn't convert cookies from `getCookies()` method in `RestAssuredResponseConverter`). My reasoning for that was that cookies from a `Set-Cookie` header carry some additional information apart from the name and a value (like domain, HttpOnly flag, etc.) and it would be better to just pass them in `RestAssuredResponseConverter`.
Now that I'm thinking about it, I'm getting more convinced that maybe it would be better to have this information structured in a more object-oriented fashion and create a `ResponseCookie` class (similar to the `RequestCookie` you've introduced).

I'll update this PR accordingly when we reach a decision (I see that I need to do a rebase anyway).

I probably should have mentioned this problem in the PR description, sorry about that.

wilkinsona:
REST Assured and Jersey's test client (for which there's third-party REST Docs support) , but set the `Set-Cookie` header automatically. It'd be nice of MockMvc wasn't the odd one out. MockMvc is also a bit inconsistent on its own: if you use a setter to configure the content type of a request or response it automatically sets the `Content-Type` header as well. I've opened [SPR-15225](https://jira.spring.io/browse/SPR-15225) to see if removing that inconsistency is a possibility. The outcome of that issue will probably tell us what is the right thing to do here.

wilkinsona:
It looks like SPR-15225 is going to be fixed, but only in 5.0. I think that means we should proceed with setting the `Set-Cookie` header for MockMvc but only if it hasn't already been set. That should allow REST Docs to work with both Spring Framework 4.3 and 5.0.

tkopczynski:
@wilkinsona I updated the PR. Is it what you had in mind?

wilkinsona:
Yes, thanks. This looks good. We talked above about making the cookies in the response available from `OperationResponse` but I think that can wait until #306.

wilkinsona:
Thanks again, @tkopczynski. This is now in master.

tkopczynski:
Great! Thanks @wilkinsona for merging this

PR #343, opened 2017-01-30
Add spring-auto-restdocs to third-party extensions

fbenz:


wilkinsona:
Thanks for the PR, @fbenz. The change is now in master.

Issue #323, opened 2016-10-27
Provide a custom Asciidoctor macro that reduces the boilerplate required to include snippets


wilkinsona:
The basic idea is that rather than using an `include` for each individual snippet, you can use a single macro to include all of the snippets for a particular operation that's been documented. For added flexibility, the macro could use attributes to control which snippets are included.

I prototyped this a while ago as a custom target for the `include` macro: https://github.com/wilkinsona/springrestdocs/commit/6657af0d590be2673f05d7ecd8f1a91244751edc. I think this could be further refined so that it becomes something like `operation::{snippets}/operation-name`.

capncodewash:
This would be very useful.

meistermeier:
@wilkinsona can you take a look at the link? I get a 404. If you don't mind I could take care of this (if we exclude the "automagical" path detection ;) )

meistermeier:
Now I finally got some time to think about this issue and playing around with some possible solutions. I don't think that a "one-include-includes-all" approach fits the rest-docs use-case but introducing categories / operations groups (no better naming found) additional to explicit named snippets might be a useful option.

    include::restdocs:index-example[category='response']

would be the same as (just as an example)

    include::restdocs:index-example[snippets='response-fields, curl-request, http-response']

What is your opinion?

wilkinsona:
I think it could be a useful option, but I'd prefer to tackle it separately and use the `snippets` attribute as a starting point.

meistermeier:
In contrast to the initial idea to use a block macro processor to provide the possibility to use `operation::{snippets}/operation-name` I kept using the include processor. The problem with converting raw input manually and missing ouput/backend format information mentioned in [gh-297 ](https://github.com/spring-projects/spring-restdocs/issues/297#issuecomment-256116925) still exists in asciidoctorj.

wilkinsona:
Ideally, I'd like to implement this as a block macro. I spoke to @mojavelinux about it a few months ago and Dan thought it should be possible but I never figured out how to do it. I thought I'd realised it was possible using the Ruby-based API but while trying to refresh my memory today I think that may have just been wishful thinking.

@mojavelinux do you have any cycles available to help us out here? In a nutshell, I'd like to be able to provide a block macro (or anything else that's similarly concise) that produces [these lines](https://github.com/spring-projects/spring-restdocs/blame/master/samples/rest-notes-spring-hateoas/src/docs/asciidoc/api-guide.adoc#L102-L120) automatically.

[...]

PR #346, opened 2017-02-08
Provide custom Asciidoctor include processor

meistermeier:
as discussed in gh-323 without categories but with level attribute


meistermeier:
No clue why the maven step fails in travis and I am aware that there should be an additional rebase afterwards.

wilkinsona:
As discussed in #323, it looks like we can do this with a block macro processor written in Ruby so I'm going to close this one. Sorry for the wasted effort, @meistermeier.

PR #348, opened 2017-02-12
Line breaks in cURL and HTTPie snippets

tkopczynski:
This PR is adding a straightforward implementation of line breaks in cURL and HTTPie snippets. It doesn't add any configuration parameter to disable them.

@wilkinsona What do you think? Is it enough or you would like to see some additional features in here?

Closes gh-260

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CommandFormatter.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CommandFormatter.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2014-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.restdocs.cli;
+
+import java.util.List;
+
+/**
+ * Formatter for {@link CurlRequestSnippet} and {@link HttpieRequestSnippet}.
+ * Its purpose is to format a command snippet from a list of its parts represented
+ * as {@code String}s.
+ *
+ * @author Tomasz Kopczynski
+ */
+public interface CommandFormatter {
+
+	/**
+	 * Formats a list of {@code String}s into a single {@code String}.
+	 *
+	 * @param elements A list of {@code String}s to be formatted
+	 * @return A list of {@code String}s formatted as one {@code String}
+	 */
+	String format(List<String> elements);
+}
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2014-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.restdocs.cli;
+
+import java.util.List;
+
+import org.springframework.util.CollectionUtils;
+
+/**
+ * {@link CommandFormatter} which concatenates commands with a given {@code separator}.
+ *
+ * @author Tomasz Kopczynski
+ */
+final class ConcatenatingCommandFormatter implements CommandFormatter {
+
+	private String separator;
+
+	ConcatenatingCommandFormatter(String separator) {
+		this.separator = separator;
+	}
+
+	/**
+	 * Concatenates a list of {@code String}s with a specified separator.
+	 *
+	 * @param elements A list of {@code String}s to be concatenated
+	 * @return Concatenated list of {@code String}s as one {@code String}
+	 */
+	@Override
+	public String format(List<String> elements) {
+		if (CollectionUtils.isEmpty(elements)) {
+			return "";
+		}
+
+		StringBuilder result = new StringBuilder();
+
+		for (String element : elements) {
+			result.append(String.format(this.separator));
+			result.append(element);
+		}
+
+		return result.toString();
+	}
+}
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
@@ -16,8 +16,7 @@
 
 package org.springframework.restdocs.cli;
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -31,6 +30,7 @@ import org.springframework.restdocs.operation.Parameters;
 import org.springframework.restdocs.operation.RequestCookie;
 import org.springframework.restdocs.snippet.Snippet;
 import org.springframework.restdocs.snippet.TemplatedSnippet;
+import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -45,11 +45,23 @@ import org.springframework.util.StringUtils;
  */
 public class CurlRequestSnippet extends TemplatedSnippet {
 
+	private final CommandFormatter commandFormatter;
+
 	/**
 	 * Creates a new {@code CurlRequestSnippet} with no additional attributes.
 	 */
+	@Deprecated
 	protected CurlRequestSnippet() {
-		this(null);
+		this(null, null);
+	}
+
+	/**
+	 * Creates a new {@code CurlRequestSnippet} with a given {@link CommandFormatter}.
+	 *
+	 * @param commandFormatter The formatter for generating the snippet
+	 */
+	protected CurlRequestSnippet(CommandFormatter commandFormatter) {
+		this(null, commandFormatter);
 	}
 
 	/**
@@ -58,8 +70,24 @@ public class CurlRequestSnippet extends TemplatedSnippet {
 	 *
 	 * @param attributes The additional attributes
 	 */
+	@Deprecated
 	protected CurlRequestSnippet(Map<String, Object> attributes) {
+		this(attributes, null);
+	}
+
+	/**
+	 * Creates a new {@code CurlRequestSnippet} with the given additional
+	 * {@code attributes} that will be included in the model during template rendering
+	 * and the given {@link CommandFormatter}.
+	 *
+	 * @param attributes The additional attributes
+	 * @param commandFormatter The formatter for generating the snippet
+	 */
+	protected CurlRequestSnippet(Map<String, Object> attributes, CommandFormatter commandFormatter) {
 		super("curl-request", attributes);
+
+		Assert.notNull(commandFormatter, "Command formatter must be set");
+		this.commandFormatter = commandFormatter;
 	}
 
 	@Override
@@ -87,21 +115,25 @@ public class CurlRequestSnippet extends TemplatedSnippet {
 	}
 
 	private String getOptions(Operation operation) {
-		StringWriter command = new StringWriter();
-		PrintWriter printer = new PrintWriter(command);
-		writeIncludeHeadersInOutputOption(printer);
-		CliOperationRequest request = new CliOperationRequest(operation.getRequest());
-		writeUserOptionIfNecessary(request, printer);
-		writeHttpMethodIfNecessary(request, printer);
-		writeHeaders(request, printer);
-		writeCookies(request, printer);
-		writePartsIfNecessary(request, printer);
-		writeContent(request, printer);
+		StringBuilder builder = new StringBuilder();
+		writeIncludeHeadersInOutputOption(builder);
 
-		return command.toString();
+		CliOperationRequest request = new CliOperationRequest(operation.getRequest());
+		writeUserOptionIfNecessary(request, builder);
+		writeHttpMethodIfNecessary(request, builder);
+
+		List<String> additionaLines = new ArrayList<>();
+		writeHeaders(request, additionaLines);
+		writeCookies(request, additionaLines);
+		writePartsIfNecessary(request, additionaLines);
+		writeContent(request, additionaLines);
+
+		builder.append(this.commandFormatter.format(additionaLines));
+
+		return builder.toString();
 	}
 
-	private void writeCookies(CliOperationRequest request, PrintWriter printer) {
+	private void writeCookies(CliOperationRequest request, List<String> lines) {
 		if (!CollectionUtils.isEmpty(request.getCookies())) {
 			StringBuilder cookiesBuilder = new StringBuilder();
 			for (RequestCookie cookie : request.getCookies()) {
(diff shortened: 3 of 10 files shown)


wilkinsona:
Thanks for the PR. This looks good; I really like the simplicity of it. I do wonder if it might benefit from being hidden behind an abstraction though.

When I was looking at #261 I had something like a `CommandFormatter` in mind that would take a list of Strings, each of which should be kept on one line. The formatter could then either concatenate them with ` ` separators, with `\\%n` separators, or whatever someone wants to implement. A particular formatter could be used by default, and could be configured using `SnippetConfigurer` or by passing one into new overloaded methods on `CliDocumentation`.

tkopczynski:
@wilkinsona Thanks!

As for your additional remarks: I was thinking about injecting a `CommandFormatter` object to both `CurlRequestSnippet` and `HttpieRequestSnippet` and using it for printing the output. But then I realized that it's not that simple. 
Let's look at the example from the initial issue:

```
$ curl 'http://localhost/foo' -i -X POST \
   -H 'X-Header-One: ONE' \
   -H 'X-Header-Two: TWO' \
   -d 'a=alpha&b=bravo'
```

Some `curl` parameters are printed after a newline and some are not (`-i -X POST`). This means that there are actually two different separators in use here. I see two possible options for dealing with this:

1. Injecting 2 instances of `CommandFormatter` and passing parameters to one or the other
2. Constructing a `CommandFormatter` object by passing two separators to it and then maintaining both parameter lists inside it.

Option 2 seems to be better for me but I'd like to reach a decision before implementing this. What do you think?

wilkinsona:
I wonder if we could keep it simple and have the curl snippet be somewhat opinionated about where line breaks could go. It could pass the following to the formatter:

- `curl 'http://localhost/foo' -i -X POST`
- `-H 'X-Header-One: ONE'`
- `-H 'X-Header-Two: TWO'`
- `-d 'a=alpha&b=bravo'`

The formatter would treat each of those 4 items as a String that should all stay on one line. The formatter is then free to join those four strings in whatever way it deems appropriate.

tkopczynski:
@wilkinsona I have created the `CommandFormatter` class and applied it to the curl and HTTPie snippets. What do you think about it?

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CommandFormatter.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Formatter for {@link CurlRequestSnippet} and {@link HttpieRequestSnippet}.
+ * Its purpose is to create a command snippet with a selected separator.
+ *
+ * @author Tomasz Kopczynski
+ */
+public final class CommandFormatter {
```
wilkinsona: Could you make this an interface, please? There could be a single implementation provided out of the box. Something like `ConcatenatingCommandFormatter` that takes a separator as a constructor argument.

tkopczynski: OK, I think I managed to introduce the `ConcatenatingCommandFormatter` class as you proposed.

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
```diff
@@ -45,6 +44,8 @@
  */
 public class CurlRequestSnippet extends TemplatedSnippet {
 
+	private CommandFormatter commandFormatter = ConcatenatingCommandFormatter.newlineSeparator();
+
 	/**
 	 * Creates a new {@code CurlRequestSnippet} with no additional attributes.
 	 */
```
wilkinsona: This constructor should be deprecated and a new constructor that takes a `CommandFormatter` introduced

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/HttpieRequestSnippet.java
```diff
@@ -46,6 +47,8 @@
  */
 public class HttpieRequestSnippet extends TemplatedSnippet {
 
+	private CommandFormatter commandFormatter = ConcatenatingCommandFormatter.newlineSeparator();
+
 	/**
 	 * Creates a new {@code HttpieRequestSnippet} with no additional attributes.
 	 */
```
wilkinsona: This constructor should be deprecated and a new constructor that takes a CommandFormatter introduced

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Concatenates a list of {@code String}s with a specified separator.
+	 *
+	 * @param elements A list of {@code String}s to be concatenated
+	 * @return Concatenated list of {@code String}s as one {@code String}
+	 */
+	@Override
+	public String format(List<String> elements) {
+		if (elements == null || elements.isEmpty()) {
```
wilkinsona: You could use `CollectionUtils.isEmpty()` here

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * {@link CommandFormatter} which concatenates commands with a given {@code separator}.
+ *
+ * @author Tomasz Kopczynski
+ */
+public final class ConcatenatingCommandFormatter implements CommandFormatter {
```
wilkinsona: I think I'd prefer it if this were package-private

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+
+	private String separator;
+
+	/**
+	 * Creates a new {@code ConcatenatingCommandFormatter} with a newline character as a separator.
+	 * @return {@code ConcatenatingCommandFormatter} with a newline character as a separator.
+	 */
+	public static ConcatenatingCommandFormatter newlineSeparator() {
```
wilkinsona: I think this static method would be better on `CliDocumentation`. I'd be tempted to call it `multiLine` or `multiLineFormat`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+		return new ConcatenatingCommandFormatter(" \\%n ");
+	}
+
+	/**
+	 * Creates a new {@code ConcatenatingCommandFormatter} with a space character as a separator.
+	 * @return {@code ConcatenatingCommandFormatter} with a space character as a separator.
+	 */
+	public static ConcatenatingCommandFormatter spaceSeparator() {
```
wilkinsona: I think this static method would be better on `CliDocumentation`. I'd be tempted to call it `singleLine` or `singleLineFormat`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+		return new ConcatenatingCommandFormatter(" ");
+	}
+
+	/**
+	 * Creates a new {@code ConcatenatingCommandFormatter} with a custom character as a separator.
+	 *
+	 * @param separator A separator to be used for concatenating
+	 * @return {@code ConcatenatingCommandFormatter} with a custom character as a separator.
+	 */
+	public static ConcatenatingCommandFormatter customSeparator(String separator) {
```
wilkinsona: I'm not sure we need this

spring-restdocs-core/src/test/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatterTests.java
```diff
@@ -0,0 +1,61 @@
...
+public class ConcatenatingCommandFormatterTests {
+
+	private CommandFormatter newlineCommandFormatter = ConcatenatingCommandFormatter.newlineSeparator();
+	private CommandFormatter spaceCommandFormatter = ConcatenatingCommandFormatter.spaceSeparator();
+	private CommandFormatter customCommandFormatter = ConcatenatingCommandFormatter.customSeparator("-");
+
+	private static final String STR = "test";
+
+	@Test
+	public void noElementsTest() {
```
wilkinsona: If the static methods are moved, this test could be simplified to just check that no elements results in an empty string. It would be good to have a similar test for `null` too.

spring-restdocs-core/src/test/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatterTests.java
```diff
@@ -0,0 +1,61 @@
...
+		assertThat(this.newlineCommandFormatter.format(Collections.<String>emptyList()), is(equalTo("")));
+		assertThat(this.spaceCommandFormatter.format(Collections.<String>emptyList()), is(equalTo("")));
+		assertThat(this.customCommandFormatter.format(Collections.<String>emptyList()), is(equalTo("")));
+
+	}
+
+	@Test
+	public void singleElementTest() {
```
wilkinsona: I think this could be simplified to just test against a single formatter with whatever separator too

spring-restdocs-core/src/test/java/org/springframework/restdocs/cli/ConcatenatingCommandFormatterTests.java
```diff
@@ -0,0 +1,61 @@
...
+		assertThat(this.newlineCommandFormatter.format(STR), is(equalTo(String.format(" \\%n %s", STR))));
+		assertThat(this.spaceCommandFormatter.format(STR), is(equalTo(String.format(" %s", STR))));
+		assertThat(this.customCommandFormatter.format(STR), is(equalTo(String.format("-%s", STR))));
+	}
+
+	@Test
+	public void twoElementsTest() {
```
wilkinsona: I think this could be simplified to just test against a single formatter with whatever separator too

spring-restdocs-core/src/test/java/org/springframework/restdocs/cli/HttpieRequestSnippetTests.java
```diff
@@ -149,7 +149,7 @@ public void postRequestWithQueryStringWithNoValue() throws IOException {
 	@Test
 	public void postRequestWithOneParameter() throws IOException {
 		this.snippets.expectHttpieRequest().withContents(codeBlock("bash")
-				.content("$ http --form POST 'http://localhost/foo' 'k1=v1'"));
+				.content(String.format("$ http --form POST 'http://localhost/foo' \\%n 'k1=v1'")));
```
wilkinsona: I'd prefer that the existing tests were largely unaffected by the introduction of the CommandFormatter. That could be done by leaving the content assertions as they were, but creating the snippet with a single-line formatter.

spring-restdocs-mockmvc/src/test/java/org/springframework/restdocs/mockmvc/MockMvcRestDocumentationIntegrationTests.java
```diff
@@ -157,8 +159,24 @@ public void curlSnippetWithContent() throws Exception {
...
 						"build/generated-snippets/curl-snippet-with-content/curl-request.adoc"),
 				is(snippet(asciidoctor()).withContents(codeBlock(asciidoctor(), "bash")
-						.content("$ curl " + "'http://localhost:8080/' -i -X POST "
-								+ "-H 'Accept: application/json' -d 'content'"))));
+						.content(String.format("$ curl " + "'http://localhost:8080/' -i -X POST "
+								+ "\\%n -H 'Accept: application/json' \\%n -d 'content'")))));
+	}
+
+	@Test
+	public void curlSnippetWithContentSpaceSeparator() throws Exception {
```
wilkinsona: It's good to have tests for the curl snippet with different command formatters, but there's nothing MockMvc-specific about it so I don't think they belong here.

spring-restdocs-mockmvc/src/test/java/org/springframework/restdocs/mockmvc/MockMvcRestDocumentationIntegrationTests.java
```diff
@@ -221,8 +239,24 @@ public void httpieSnippetWithContent() throws Exception {
...
 						"build/generated-snippets/httpie-snippet-with-content/httpie-request.adoc"),
 				is(snippet(asciidoctor()).withContents(codeBlock(asciidoctor(), "bash")
-						.content("$ echo 'content' | http POST 'http://localhost:8080/'"
-								+ " 'Accept:application/json'"))));
+						.content(String.format("$ echo 'content' | http POST 'http://localhost:8080/'"
+								+ " \\%n 'Accept:application/json'")))));
+	}
+
+	@Test
+	public void httpieSnippetWithContentSpaceSeparator() throws Exception {
```
wilkinsona: It's good to have tests for the HTTPie snippet with different command formatters, but there's nothing MockMvc-specific about it so I don't think they belong here.

spring-restdocs-restassured/src/test/java/org/springframework/restdocs/restassured/RestAssuredRestDocumentationIntegrationTests.java
```diff
@@ -113,10 +115,29 @@ public void curlSnippetWithContent() throws Exception {
...
 				is(snippet(asciidoctor()).withContents(codeBlock(asciidoctor(), "bash")
-						.content("$ curl 'http://localhost:" + this.port + "/' -i "
+						.content(String.format("$ curl 'http://localhost:" + this.port + "/' -i "
+								+ "-X POST \\%n -H 'Accept: application/json' "
+								+ "\\%n -H 'Content-Type: " + contentType + "' "
+								+ "\\%n -d 'content'")))));
+	}
+
+	@Test
+	public void curlSnippetWithContentSpaceSeparator() throws Exception {
```
wilkinsona: It's good to have tests for the curl snippet with different command formatters, but there's nothing REST Assured-specific about it so I don't think they belong here.

spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
tkopczynski: Just a quick question about this one: deprecating this constructor would mean that I would have to touch every test in `CurlRequestSnippetTests` to switch to a different constructor. Is that ok for you?

wilkinsona: I was torn on this earlier when I was reviewing the change but, while annoying, I think it's probably the right thing to do. It's what I was getting at when I said "but creating the snippet with a single-line formatter" in my comment on the tests below.

My motivation was that I wanted the opinion on what the default formatter should be to be in a single place. `CliDocumentation` seemed to make sense as it can then pass the default formatter into `CurlRequestSnippet` and `HttpieRequestSnippet`. 

I was also envisaging a further change where the default moves to `SnippetConfigurer` and it'll apply it to the default `CurlRequestSnippet` and `HttpieRequestSnippet` that it creates. The would lead to the deprecation of `curlRequest()` and `httpieRequest()` on `CliDocumentation` as well.

As I said, I was torn on this so if you don't think the reasoning above is sound I'd really like to hear any counter arguments.

tkopczynski: OK, I understand now. I'm also in favour of putting the default formatter in one place, I just wasn't sure where should it be. After looking at the code again I think that `CliDocumentation` should be a good enough place to put it there.

But, to have it done in a clean way I would propose the following changes:

- `CurlRequestSnippet`: create a copy of all constructors with an additional `commandFormatter` parameter and deprecate the old ones
- `CliDocumentation`: create additional `curlRequest` method that takes `attributes` and `commandFormatter` parameters. Additionally, change all existing `CurlRequestSnippet` constructors so that they will receive the default formatter.
- `CurlRequestSnippetTests`: add the additional formatter parameter to each `CurlRequestSnippet` constructor or call the appropriate `CliDocumentation` method. The former option feels better since these are the `CurlRequestSnippet` tests.

I do understand that this is more or less the same that you have proposed in the first place but now it's written in one place and it's easier for me to see the whole picture.

wilkinsona: Sounds good to me. Thanks very much.

tkopczynski:
@wilkinsona I've updated the PR as we discussed, please take a look.

wilkinsona:
@tkopczynski Thanks

wilkinsona:
Thanks again for the PR. This change is now in master.

tkopczynski:
Great, thanks for the review @wilkinsona

PR #350, opened 2017-02-18
Set SnippetConfiguration class to public

jmisur:
In our project https://github.com/ScaCap/spring-auto-restdocs we sometimes write multi-line values inside tables in snippets. We need to discover the templateFormat which will be used during writing such snippet, so that we can output proper newline character: ` +\n` in AsciiDoc and `<br>` in Markdown.
Therefore we need this class to be public, so we can extract this information by getting it from `org.springframework.restdocs.operation.Operation` attributes.

If this is not acceptable, we need another way of getting template format information from the Operation to use it inside the snippet.


wilkinsona:
Sorry, but I don't want to make that class public. Can you please open an issue and we can explore another solution. I suspect that a Mustache lambda will be the answer. REST Docs already has one for escaping `|` characters in Asciidoctor tables.

Issue #347, opened 2017-02-09
Provide http(s) default port method


meistermeier:
It feels like a magic number to use `80` or `443` in the `UriConfigurer#withPort` method to suppress the port output given the right scheme.

There should be a method to set this magic number without user getting to see or having to use it like `withHttpPort` or `withHttpsPort`.

I know it is a small change but I think the right way is to discuss it before opening a PR.

wilkinsona:
How about `withHttp()` and `withHttps()` methods that set both the scheme and the port? They'd be shorthand for `withScheme("http").withPort(80)` and `withScheme("https").withPort(443)` respectively.

meistermeier:
I was thinking the same but was not sure if this might sound ambiguous (just the scheme or both) for the user. On the other hand this should be solved by a good documentation.

wilkinsona:
That's certainly a risk. Another option would be to deprecate `withScheme(String)` and `withPort(int)` in favour of:

- `withHttp()`
- `withHttp(int)`
- `withHttps()`
- `withHttps(int)`

The methods would set the scheme and the port. Methods that don't take an `int` set the port to the default for the scheme.

meistermeier:
That might be a better idea. Just one little thing: it _could_ be confusing to have default port 8080 and switching it to 80 by using the `withHttp()` method. But in my opinion it is much better to provide this feature with a short and clear name and document it than having to pass numbers and strings 😁 to the configuration.

wilkinsona:
> it could be confusing to have default port 8080 and switching it to 80 by using the withHttp() method

Hmm, I think that would be confusing. I think that puts that particular idea at the bottom of the list for me.

Another option would be for `UriConfigurer` to provide constants for the standard HTTP and HTTPS ports that can be passed to the existing `withPort(int)` method. I think that might be my preferred option at the moment as it doesn't introduce any new methods to the API and therefore hopefully doesn't introduce any possible confusion either.

meistermeier:
This simple improvement just points out how complicated it can be to introduce such a little change. I can understand that the pro and cons put this on the bottom on the list. (deleting local branch :grin: )

PR #353, opened 2017-02-20
Add new methods for UriConfigurer

eddumelendez:
This commit adds new methods as a shorthand for the existing ones.
`withHttp` and `withHttps` allows to override the default port.

See gh-347


wilkinsona:
Thanks for the PR, @eddumelendez. Unfortunately, the discussion in #347 didn't reach a conclusion. All of the proposed options, including the one implemented here, had some downsides. I am not convinced that the benefits of any of them outweigh the downsides so I'm going to close this.

PR #354, opened 2017-02-25
Provide a custom Asciidoctor include macro

meistermeier:
uses a ruby based solution to add snippet blocks to the parent block element

Fixes gh-323

### DIFF
--- a/spring-restdocs-asciidoctor/src/main/java/org/springframework/restdocs/asciidoctor/RestDocsExtensionRegistry.java
+++ b/spring-restdocs-asciidoctor/src/main/java/org/springframework/restdocs/asciidoctor/RestDocsExtensionRegistry.java
@@ -17,6 +17,8 @@
 package org.springframework.restdocs.asciidoctor;
 
 import org.asciidoctor.Asciidoctor;
+import org.asciidoctor.extension.JavaExtensionRegistry;
+import org.asciidoctor.extension.RubyExtensionRegistry;
 import org.asciidoctor.extension.spi.ExtensionRegistry;
 
 /**
@@ -28,8 +30,15 @@ public final class RestDocsExtensionRegistry implements ExtensionRegistry {
 
 	@Override
 	public void register(Asciidoctor asciidoctor) {
-		asciidoctor.javaExtensionRegistry()
-				.preprocessor(new DefaultAttributesPreprocessor());
+		JavaExtensionRegistry registry = asciidoctor.javaExtensionRegistry();
+		registry.preprocessor(new DefaultAttributesPreprocessor());
+
+		RubyExtensionRegistry rubyExtensionRegistry = asciidoctor.rubyExtensionRegistry();
+		rubyExtensionRegistry
+				.loadClass(RestDocsExtensionRegistry.class
+						.getResourceAsStream("/extensions/operation_block_macro.rb"))
+				.blockMacro("operation", "OperationBlockMacro");
+
 	}
 
 }
--- a/samples/rest-notes-spring-hateoas/build.gradle
+++ b/samples/rest-notes-spring-hateoas/build.gradle
@@ -4,6 +4,8 @@ buildscript {
 	}
 	dependencies {
 		classpath 'org.springframework.boot:spring-boot-gradle-plugin:1.4.1.RELEASE'
+		classpath 'org.asciidoctor:asciidoctorj:1.5.4'
+		classpath "org.springframework.restdocs:spring-restdocs-asciidoctor:1.2.0.BUILD-SNAPSHOT"
 	}
 }
 
--- a/samples/rest-notes-spring-hateoas/build.gradle
+++ b/samples/rest-notes-spring-hateoas/build.gradle
@@ -4,8 +4,6 @@ buildscript {
 	}
 	dependencies {
 		classpath 'org.springframework.boot:spring-boot-gradle-plugin:1.4.1.RELEASE'
-		classpath 'org.asciidoctor:asciidoctorj:1.5.4'
-		classpath "org.springframework.restdocs:spring-restdocs-asciidoctor:1.2.0.BUILD-SNAPSHOT"
 	}
 }
 
--- a/samples/rest-notes-spring-hateoas/src/docs/asciidoc/api-guide.adoc
+++ b/samples/rest-notes-spring-hateoas/src/docs/asciidoc/api-guide.adoc
@@ -10,6 +10,8 @@ Andy Wilkinson;
 [[overview]]
 = Overview
 
+operation::index-example[snippets='response-fields,http-response', level=2]
+
 [[overview-http-verbs]]
 == HTTP verbs
 
--- a/samples/rest-notes-spring-hateoas/src/docs/asciidoc/api-guide.adoc
+++ b/samples/rest-notes-spring-hateoas/src/docs/asciidoc/api-guide.adoc
@@ -10,8 +10,6 @@ Andy Wilkinson;
 [[overview]]
 = Overview
 
-operation::index-example[snippets='response-fields,http-response', level=2]
-
 [[overview-http-verbs]]
 == HTTP verbs
 
--- a/spring-restdocs-asciidoctor/build.gradle
+++ b/spring-restdocs-asciidoctor/build.gradle
@@ -5,3 +5,12 @@ dependencies {
 	testCompile 'junit:junit'
 	testCompile 'org.asciidoctor:asciidoctorj'
 }
+
+task copyTestSnippets(type: Copy) {
+	from 'src/test/resources/some-operation'
+	into 'build/generated-snippets/some-operation'
+}
+
+test {
+	dependsOn copyTestSnippets
+}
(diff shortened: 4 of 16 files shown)


spring-restdocs-asciidoctor/src/main/java/org/springframework/restdocs/asciidoctor/RestDocsExtensionRegistry.java
```diff
@@ -28,8 +30,15 @@
 
 	@Override
 	public void register(Asciidoctor asciidoctor) {
-		asciidoctor.javaExtensionRegistry()
-				.preprocessor(new DefaultAttributesPreprocessor());
+		final JavaExtensionRegistry registry = asciidoctor.javaExtensionRegistry();
```
wilkinsona: REST Docs' code style is to only use `final` when it's required

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
+require 'asciidoctor/extensions'
+require 'stringio'
+
+# Spring REST Docs block macro to import multiple snippet of an operation at once
+#
+# Usage
+#
+#   srd::operation-name[snippets='snippet-name1,snippet-name2', level=<indentation level>]
+#
+class RestDocsBlockMacro < Asciidoctor::Extensions::BlockMacroProcessor
```
wilkinsona: I think this should be called `OperationBlockMacro` and the file itself should be called `operation_block_macro.rb`

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
...
+               :'response-fields' => 'Response fields',
+               :'links' => 'Links'
+    }
+  end
+
+  def process(parent, reader, attrs)
+    doc = parent.document
+    snippet_dir = doc.attributes['snippets']
+    snippets = snippets_from_attributes(attrs)
+    section_level = section_level_from_attributes(attrs)
```
wilkinsona: I think we could use `parent.level` to figure this out automatically

spring-restdocs-asciidoctor/src/main/java/org/springframework/restdocs/asciidoctor/RestDocsExtensionRegistry.java
```diff
@@ -28,8 +30,15 @@
...
+		registry.preprocessor(new DefaultAttributesPreprocessor());
+
+		RubyExtensionRegistry rubyExtensionRegistry = asciidoctor.rubyExtensionRegistry();
+		rubyExtensionRegistry
+				.loadClass(RestDocsExtensionRegistry.class
+						.getResourceAsStream("/extensions/srd-include-block-macro.rb"))
+				.blockMacro("srd", "RestDocsBlockMacro");
```
wilkinsona: I don't think the macro should be called `srd` as that would seem to imply the REST Docs will only ever have a single block macro. There doesn't seem to be a tradition of trying to namespace macro names in Asciidoctor so I'm tempted to go with `operation` instead.

spring-restdocs-asciidoctor/src/test/java/org/springframework/restdocs/asciidoctor/RestDocsIncludeBlockMacroTest.java
```diff
@@ -0,0 +1,165 @@
...
+				"</div>";
+
+		String result = this.asciidoctor.convert(
+				"srd::some-operation[snippets='curl-request']", this.options);
+
+		assertThat(result, equalTo(expectedContent));
+	}
+
+	@Test
+	public void useMacroWithoutSnippetAddsNothing() {
```
wilkinsona: I'm intrigued by this test. What was the reasoning behind adding nothing? I would expect this to either produce an error or to include all of the snippets for a particular operation. My preference is for the latter but I'm not set on that.

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+  end
+
+  # if no section indentation level is provided, default to 4
+  def section_level_from_attributes(attrs)
+    (attrs['level'].to_i if attrs['level']) || 4
+  end
+
+  def snippets_from_attributes(attrs)
+    (attrs['snippets'].split(',') if attrs['snippets']) || []
```
wilkinsona: I think using [`fetch`](https://ruby-doc.org/core-1.9.3/Hash.html#method-i-fetch) would make this read a bit better:

```ruby
attrs.fetch('snippets', '').split(',')
```

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  def append_snippet_block(content, snippet, params)
+    write_title content, snippet, params[:section_level]
+    write_content content, snippet, params
+  end
+
+  def write_content(content, snippet, params)
+    asciidoc_extension = '.adoc'
+    snippet_path = params[:snippet_dir].to_s + File::SEPARATOR +
```
wilkinsona: I think using `File.join` and some string interpolation would improve the readability here. Something like:

```ruby
File.join(params[:snippet_dir], params[:operation], "#{snippet}.adoc")
```

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
...
+        params[:operation] + File::SEPARATOR +
+        snippet + asciidoc_extension
+
+    content.puts File.readlines(snippet_path).join
+
+  rescue Errno::ENOENT
+    content.puts "WARNING: snippet not found: " + snippet_path
```
wilkinsona: I'd use string interpolation here too

spring-restdocs-asciidoctor/src/main/resources/extensions/srd-include-block-macro.rb
```diff
@@ -0,0 +1,103 @@
...
+    add_new_line content
+  end
+
+  def write_title(content, snippet, level)
+    section_level = '=' * level
+    content.puts section_level + ' ' + title(snippet)
```
wilkinsona: I'd use string interpolation here too

spring-restdocs-asciidoctor/src/test/java/org/springframework/restdocs/asciidoctor/RestDocsIncludeBlockMacroTest.java
```diff
@@ -0,0 +1,165 @@
...
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for ruby based rest docs block macro
+ * Because there is no java implementation (yet)
+ * we can only test the behaviour when rendering
+ *
+ * @author Gerrit Meier
+ */
+public class RestDocsIncludeBlockMacroTest {
```
wilkinsona: This should be named `…Tests` to match the other tests classes

spring-restdocs-asciidoctor/src/test/java/org/springframework/restdocs/asciidoctor/RestDocsIncludeBlockMacroTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+	private Attributes getAttributes() {
+		Attributes attributes = new Attributes();
+		attributes.setAttribute("projectdir", new File(".").getAbsolutePath());
+		return attributes;
+	}
+
+	@Test
+	public void includeSnippetWithLevel() {
+		String expectedContent = "<div class=\"sect1\">\n" +
```
wilkinsona: My preference would be to store the expected content in files in `src/test/resources`

meistermeier:
Thanks for the feedback. I think I can do the changes but this may take until the next weekend.

spring-restdocs-asciidoctor/src/test/java/org/springframework/restdocs/asciidoctor/RestDocsIncludeBlockMacroTest.java
meistermeier: Now it adds every exiting snippet in the operation folder.

meistermeier:
I have updated the code and rebased it (so that the code comments above are broken sorry for that 🙂  ). 
Major changes:
* the include without any snippets will include every snippet in the operation directory
* the level of the include will be derived from the parent block level

wilkinsona:
Thanks, @meistermeier. Did you miss some files when you made the update? The new tests are failing for me because they can't find the `build/generated-snippets/some-operation` directory

meistermeier:
There went something completely wrong with the update...so I went down in the code dungeon and found some commits with loose endings ;)

wilkinsona:
@meistermeier I've merged this into master. Thank you very much for the contribution.

@mojavelinux We couldn't have done this without your help. Thank you very much.

PR #379, opened 2017-04-09
Extend the default snippets list in documentation

tkopczynski:
Adding two additional snippets to the default snippets list in documentation


wilkinsona:
Thank you

Issue #382, opened 2017-04-17
Example samples/rest-notes-spring-data-rest build fails


kpb:
### To Reproduce ###
clean .m2 repo. Clone project. Attempt build/package.
```sh
$ m -rf ~/.m2/repository/
$ git clone https://github.com/spring-projects/spring-restdocs.git
$ cd spring-restdocs/samples/rest-notes-spring-data-rest
$ mvn package
[INFO] ------------------------------------------------------------------------
[INFO] Building rest-notes-spring-data-rest 0.0.1-SNAPSHOT
...
[INFO] --- asciidoctor-maven-plugin:1.5.3:process-asciidoc (generate-docs) @ rest-notes-spring-data-rest ---
[WARNING] The POM for org.springframework.restdocs:spring-restdocs-asciidoctor:jar:1.2.0.BUILD-SNAPSHOT is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 11.125 s
[INFO] Finished at: 2017-04-17T12:34:57-06:00
[INFO] Final Memory: 20M/217M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc (generate-docs) on project rest-notes-spring-data-rest: Execution generate-docs of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc failed: Plugin org.asciidoctor:asciidoctor-maven-plugin:1.5.3 or one of its dependencies could not be resolved: Failure to find org.springframework.restdocs:spring-restdocs-asciidoctor:jar:1.2.0.BUILD-SNAPSHOT in https://oss.sonatype.org/content/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of sonatype-nexus-snapshots has elapsed or updates are forced -> [Help 1]
```
### Environment ###
```sh
$ java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13)
OpenJDK 64-Bit Server VM (build 25.121-b13, mixed mode)
```
```sh
$ mvn --version
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T13:39:06-06:00)
[...]

PR #383, opened 2017-04-18
Data REST sample fails to build due without a prior mvn install due to missing <pluginRepository> declaration

meistermeier:
The build of rest-notes-spring-data-rest failed because the spring-restdocs-asciidoctor artifact
is (of course) missing in the sonatype repository.
Maven does not use the repository definition to look up a dependency for a plugin but needs the
pluginRepositories configuration.

Fixes gh-#382


wilkinsona:
Thank you

Issue #393, opened 2017-04-30
Field snippets do not support XML payloads other than application/xml


bill-okara:
Hi,

First of all, I just want to sincerely say THANK YOU for creating this project. I really appreciated the flexibility and robustness it provided for doing documentation (I originally tried to upgrade our existing doc framework with Swagger/OpenApi, but that required too much refactoring of existing servlet based APIs). 

Regarding this ticket, it seems that current AbstractFieldsSnippet.getContentHandler() only selects XmlContentHandler for MediaType.APPLICATION_XML, would it makes sense to also support
- MediaType.TEXT_XML?

As stated in [RFC-3023](https://tools.ietf.org/html/rfc3023#page-5), it seems that both 'application/xml' and 'text/xml' are valid content-type for xml responses.

> If an XML document -- that is, the unprocessed, source XML document
>    -- is readable by casual users, text/xml is preferable to
>    application/xml.  MIME user agents (and web user agents) that do not
>    have explicit support for text/xml will treat it as text/plain, for
>    example, by displaying the XML MIME entity as plain text.
>    Application/xml is preferable when the XML MIME entity is unreadable
>    by casual users.
> 

Thanks!
Bill

wilkinsona:
Thank you for the kind words. Not supporting `text/xml` is an oversight on my part. `XmlContentHandler` should deal with anything that's XML (so not only `application/xml` and `text/xml` but also things like `application/rss+xml` as well).

PR #394, opened 2017-05-06
Support text/xml & application/rss+xml media types

eddumelendez:
This commit adds support to process two more content types.

See gh-393

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/AbstractFieldsSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/AbstractFieldsSnippet.java
@@ -221,7 +221,7 @@ public abstract class AbstractFieldsSnippet extends TemplatedSnippet {
 	private ContentHandler getContentHandler(byte[] content, MediaType contentType) {
 		try {
 			if (contentType != null
-					&& MediaType.APPLICATION_XML.isCompatibleWith(contentType)) {
+					&& isCompatibleWithXml(contentType)) {
 				return new XmlContentHandler(content);
 			}
 			else {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks, @eddumelendez. Rather than hard-coding some additional media types, I think it would be better to check for compatibility against `*/*+xml`. That would mean that standard media types like `application/xml`, `text/xml`, and `application/rss+xml` are all supported, as well as custom media types like `application/vnd.com.example+xml`

Would you like to update your PR to do that?

eddumelendez:
@wilkinsona Thanks for your feedback. Glad to update the PR and now I'm doing the following

```java
private boolean isCompatibleWithXml(MediaType contentType) {
    MediaType xmlMediaType = new MediaType("*", "*+xml");
    return xmlMediaType.isCompatibleWith(contentType);
}
```

but since the type is "*" even "application/json" is valid for xml. Pretty sure I misunderstood something in the process to do this. Could you provide any hint? Thanks in advance.

wilkinsona:
Nope, you haven't misunderstood. Using a check against `*/*+xml` was wishful thinking on my part. We could check for `application/xml`, `text/xml`, `application/*+xml` and `text/*+xml` but that would still miss custom XML media types as well as things like `image/svg+xml`. 

I now think that the most robust approach would be to treat the content as JSON by default. If that throws an exception then try XML. If that throws an exception then give up.

wilkinsona:
Thank you for your efforts here, @eddumelendez. There's more time pressure than usual here as 1.2.1.RELEASE is needed for Spring Boot 2.0.0.M1 so I've fixed #393 in https://github.com/spring-projects/spring-restdocs/commit/113976f2d9eb076e0f38e343629f23bea95155b9. Thanks anyway.

PR #415, opened 2017-08-14
Fix JSON field type resolution for nested array fields

petermnhk:
This PR updates the logic to determine the type of nested array correctly.

For example,
a path "[]" was correctly determined as Array,
but a path "[].a[]" was incorrectly determined as Object.

This PR determines the path "[].a[]" as Array rather than Object.


petermnhk:
Since this PR results some test failures, I recreated pull request #416.

PR #416, opened 2017-08-15
Fix JSON field type resolution for nested array fields

petermnhk:
This PR updates the logic to determine the type of nested array correctly.

For example,
a path "[]" was correctly determined as Array,
but a path "[].a[]" was incorrectly determined as Object.

This PR determines the path "[].a[]" as Array rather than Object.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldPathPayloadSubsectionExtractor.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldPathPayloadSubsectionExtractor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


wilkinsona:
Thanks very much for noticing and reporting the problem. I agree that `[].a[]` should be identified as an array. I've spent quite a bit of time reviewing the changes that I've made and I'm not convinced that it's the right approach to fixing the problem. In particular, I'm not keen on using `JsonFieldList` to identify a "special" kind of list. 

Some of the test changes also look a bit suspect to me. For example both `a.b[]` and `a[]` are "precise" paths as they identify a specific item in a single place in the payload rather than possibly matching multiple places in the payload.

Both `array` and `precise` on `JsonFieldPath` are a bit ugly and can be confusing. Judging by the comments you added to the field, I suspect you found that too. I haven't worried too much about it in the past as `JsonFieldPath` is a package-private internal, but this issue has convinced me that it's worth spending some time straightening this out properly. To that end, I'm going to open a separate issue to tackle the problem that you've identified using a different approach. Thanks anyway.

wilkinsona:
I've opened #422 to track the problem reported here. I've also found another problem that's related. I've opened #423 to track that one.

petermnhk:
Thanks for your review. I agree `array` and `precise` are confusing. I also felt the meaning of `precise` has been changed during some commits, as some of its unit tests were changed. I hope these problems will be solved soon. Thanks.

PR #430, opened 2017-10-02
Reuse ObjectMapper in FieldPathPayloadSubsectionExtractor

izeye:
This PR makes `FieldPathPayloadSubsectionExtractor` reuse `ObjectMapper`.


izeye:
I overlooked it and It won't be reused.

Sorry for the noise.

PR #431, opened 2017-10-02
Reuse ObjectMapper in FieldPathPayloadSubsectionExtractor

izeye:
This PR makes `FieldPathPayloadSubsectionExtractor` reuse `ObjectMapper`.

I originally created #430 for this but closed it because making it a field won't work. Making it `static` will work, so I reopen this. I tried to reopen #430 after force push but it looks impossible to reopen a PR after force push in GitHub.


wilkinsona:
Thank you, @izeye. I've merged this into 1.2.x and forward into master.

PR #432, opened 2017-10-04
Add linkWithSelfRel

drumonii:
Adds `linkWithSelfRel` to `HypermediaDocumentation` as a convenience method in place of `linkWithRel("self")` due to its abundance


wilkinsona:
Thanks for the PR.

I'm not sure that I like this change as it means that `HypermediaDocumentation` now has knowledge about a particular `rel`. Depending on the response format, a self link may not use `self` as its rel. I'd prefer to keep `HypermediaDocumentation` in its current, flexible form, where anything that's specific, or potentially specific, to a particular response format is handled by a `LinkExtractor`. 

Rather than adding this functionality to REST Docs, you could easily use your own convenience method.

PR #449, opened 2017-11-06
Fix name of rel

michael-simons:
I'm quite sure that the relation is usually called `self`.


wilkinsona:
Thanks very much, @michael-simons. This change is now in both 1.2.x and master.

PR #482, opened 2018-02-18
asciidocs cols seems to require quotes

bartolom:
While the [oficial asciidoctor.org reference for tables](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#tables) seems a bit ambiguous and does not use quotes for the `cols` attribute all the time, it seems that quotes are required for the example of `[cols="1,3"]`in Line 139

My setup:
spring-restdocs 2.0.0.RELEASE, Gradle plugin `"org.asciidoctor.convert" version "1.5.3`, Spring Boot 2.0.0.RC1, Java 9, Ubuntu


wilkinsona:
Thanks very much for making your first contribution to REST Docs. I've merged the change into 1.2.x and forwards into master.

PR #488, opened 2018-03-09
Curl snippet may use incorrect HTTP method for GET requests with a body

izeye:
With `-d` option, `curl` assumes its method is `POST`, so `GET` method with `-d` option is changed to `POST` method now.

This PR changes to add `-X GET` when `curl` command has `-d` option although its method is `GET`.

I didn't change to omit `-X POST` with `-d` option intentionally as it triggers to change many tests but if the change is preferred, I can update this PR to do that.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/CurlRequestSnippet.java
@@ -105,7 +105,7 @@ public class CurlRequestSnippet extends TemplatedSnippet {
 
 		CliOperationRequest request = new CliOperationRequest(operation.getRequest());
 		writeUserOptionIfNecessary(request, builder);
-		writeHttpMethodIfNecessary(request, builder);
+		writeHttpMethod(request, builder);
 
 		List<String> additionalLines = new ArrayList<>();
 		writeHeaders(request, additionalLines);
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks, @izeye. Looking at the code, I think there may be a similar problem with a multipart GET request. Do you agree?

izeye:
@wilkinsona Thanks for the quick feedback!

It looks you're right although I'm not familiar with multipart GET requests.

When checking your comment, I noticed `CurlRequestSnippetTests.multipartPostWithParameters` looks broken with `GET` method (I replaced backticks with double quotes not to break the enclosing markdown):

```
Expected: is md snippet"""bash
$ curl 'http://localhost/upload' -i -X POST -H 'Content-Type: multipart/form-data' -F 'image=@documents/images/example.png' -F 'a=apple' -F 'a=avocado' -F 'b=banana'
"""
     but: was:"""bash
$ curl 'http://localhost/upload?a=apple&a=avocado&b=banana' -i -H 'Content-Type: multipart/form-data' -F 'image=@documents/images/example.png' -F 'a=apple' -F 'a=avocado' -F 'b=banana'
"""
```

Note that the parameters have been added to the URL.

Anyway, back to this PR, I'd like to suggest to add "-X" blindly as doing so makes things simple and has no problem with these edge cases although it looks a bit verbose. If you're okay with the suggestion, I can rework this PR that way.

wilkinsona:
> It looks you're right although I'm not familiar with multipart GET requests.

Ideally, no one would be familiar with them. They're extremely unusual and really not a good idea.

> Anyway, back to this PR, I'd like to suggest to add "-X" blindly as doing so makes things simple and has no problem with these edge cases although it looks a bit verbose.

That sounds good to me.

>  If you're okay with the suggestion, I can rework this PR that way.

That'd be great, thanks.

A similar change for HTTPie is probably in order. That should probably be tackled separately though, I think.

izeye:
@wilkinsona Thanks for the feedback!

I reworked as discussed in 330ad5344c4f5bdb8541feefff7be66466ff163e.

izeye:
@wilkinsona In regards to HTTPie it looks writing HTTP method option unconditionally already although its name is not aligned with what it's doing as it's unconditional: https://github.com/spring-projects/spring-restdocs/blob/master/spring-restdocs-core/src/main/java/org/springframework/restdocs/cli/HttpieRequestSnippet.java#L149-L151

wilkinsona:
Thanks again, @izeye. These changes are now in 1.2.x and master.

PR #493, opened 2018-03-16
Move UriModifyingOperationPreprocessor to core project

joemccall86:
This class does not rely on RestAssured, and is useful for extensions to use (such as rest-template).

This may introduce breakage on existing projects that use the old package. I'm open to suggestions for how we want to handle that.


wilkinsona:
Thanks very much for making your first contribution to Spring REST Docs, @joemccall86. Your changes are now in master along with [a little bit of polish](https://github.com/spring-projects/spring-restdocs/commit/e7d2d44a0568a3b258cd788dac242d046453cd0b).

PR #508, opened 2018-05-08
Handling version when using with spring-boot

rajadilipkolli:
If we use spring boot parent property then upgrading spring boot won't cause any issue.


wilkinsona:
Closing due to the lack of the requested feedback.

Issue #299, opened 2016-09-07
Produce URI Snippet


hughwphamill:
It would be useful to be able to document a resources URI as part of documenting a given resource, this could use the custom scheme, server and port as configured in mock mvc.

It would be great if it could also include placeholders so output could something like

http://api.example.com/resources

for a collection, or 

http://api.example.com/resources/{id}

For an individual resource

Bonus points if it could also produce a path variable table that could describe path variables like {id} above.

This would be obviously useful to document consumers, but also for producers as they wouldn't have to coordinate between the request snippets produced by restdocs and 'hardcoded' URIs in the asciidoc templates.

wilkinsona:
> Bonus points if it could also produce a path variable table that could describe path variables like {id} above.

This is already possible using the [path parameters snippet](http://docs.spring.io/spring-restdocs/docs/1.1.2.RELEASE/reference/html5/#documenting-your-api-path-parameters).

> It would be useful to be able to document a resources URI as part of documenting a given resource, this could use the custom scheme, server and port as configured in mock mvc.

Generally speaking, I'm not keen on encouraging people to prominently document URIs. It leads to brittle documentation and can undo all of the hard work and benefits of adopting hypermedia. Unless there's an overwhelming number of people who would like this to be supported out of the box, I'd prefer to leave it as something that you add yourself. It's only a handful of lines:

```
public class UriSnippet extends TemplatedSnippet {

	public UriSnippet() {
		super("uri", null);
	}

	@Override
	protected Map<String, Object> createModel(Operation operation) {
		Map<String, Object> model = new HashMap<>();
		model.put("uri", operation.getRequest().getUri());
		return model;
	}
}
```

The accompanying template, named `uri.adoc` in `src/test/resources/org/springframework/restdocs/templates` would look something like this:

```
{{uri}}
```

The new snippet can then be used like this:

```
mockMvc.perform(get("/").accept(MediaType.APPLICATION_JSON))
		.andExpect(status().isOk())
		.andDo(document("uri-snippet-example", new UriSnippet()));
```

Alternatively, you could configure it as one of the [default snippets](http://docs.spring.io/spring-restdocs/docs/1.1.2.RELEASE/reference/html5/#configuration-default-snippets) so that it's always generated.

hughwphamill:
[...]

PR #516, opened 2018-06-14
Add urlTemplate model property to allow more generic request doc

romeara:
In many cases when documenting requests, it would be preferable to display the general form of a URL, as opposed to the specific form with IDs that was used to generate the documentation via MockMvc Rest Doc.

This request adds a "urlTemplate" model property to the default properties (borrowing from logic in PathParametersSnippet), so that generating using the template instead of the test path is simplified to overriding the http-request snippet in a consuming project

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/http/HttpRequestSnippet.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/http/HttpRequestSnippet.java
@@ -78,6 +78,9 @@ public class HttpRequestSnippet extends TemplatedSnippet {
 		if (urlTemplate != null) {
 			model.put("urlTemplate", urlTemplate);
 		}
+		else {
+			model.put("urlTemplate", "");
+		}
 
 		return model;
 	}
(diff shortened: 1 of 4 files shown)


romeara:
This was prompted by both my own desire for this functionality, and various discussions in existing issues. 

No default behavior is changed by this request - this just adds an option to more easily change to using the template.

I will be more than happy to make changes if desired :)

wilkinsona:
Thanks for opening your first Spring REST Docs PR.

The HTTP request snippet is supposed to contain an HTTP request as made to the server, i.e. something that you could copy and paste into telnet if you really wanted to. As such, I don't think it's appropriate for it to contain a URI template, even if you have to use a custom snippet template to get it to do so.

I was at one point completely opposed to the idea of encouraging the documentation of a URI but that's no longer the case. However, I do think it should be done in the form of a separate snippet rather than forcing it into the existing HTTP request snippet. See https://github.com/spring-projects/spring-restdocs/issues/299 for details.

If you disagree and would like the HTTP request snippet to offer the functionality, I would recommend subclassing HttpRequetSnippet in your own codebase and overriding the creation of the model.

PR #517, opened 2018-06-15
Add a snippet which documents defining properties of a request

romeara:
Add RequestUriSnippet, which by default documents the HTTP verb, URI (template), and media type properties of an HTTP request

Fixes gh-299

### DIFF
--- a/spring-restdocs-core/src/test/java/org/springframework/restdocs/request/RequestUriSnippetTests.java
+++ b/spring-restdocs-core/src/test/java/org/springframework/restdocs/request/RequestUriSnippetTests.java
@@ -71,8 +71,7 @@ public class RequestUriSnippetTests extends AbstractSnippetTests {
 				.request("http://localhost/foo").header("Alpha", "a").param("b", "bravo")
 				.build());
 	}
-
-	// TODO
+	
 	@Test
 	public void getRequestWithQueryString() throws IOException {
 		this.snippets.expectRequestUri()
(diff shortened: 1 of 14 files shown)


romeara:
@wilkinsona Is this approach what you were looking for for #299 , or was there a different direction you wanted to go? I'll be happy to adjust this to be more in-line with the rest of the library if you think it deviates in some way from the standard patterns

wilkinsona:
@romeara I'm not sure yet as other commitments have meant that I haven't had a time to review your proposal yet. It is on my to-do list though and definitely has not been forgotten.

spring-restdocs-core/src/test/java/org/springframework/restdocs/request/RequestUriSnippetTests.java
```diff
@@ -0,0 +1,306 @@
...
+				.withContents(requestUri(RequestMethod.GET, "/{a}/{b}"));
+
+		new RequestUriSnippet().document(this.operationBuilder
+				.attribute(RestDocumentationGenerator.ATTRIBUTE_NAME_URL_TEMPLATE,
+						"/{a}/{b}")
+				.request("http://localhost/foo").header("Alpha", "a").param("b", "bravo")
+				.build());
+	}
+
+	// TODO
```
jlstrater: what is the todo?

romeara: Me not cleaning up properly - it was for writing the tests, which was completed. Thanks for the catch!

huxaiphaer:
Resolve the confilcts @romeara

romeara:
Unfortunately, I still haven't been able to adjust this to the accumulated changes, and at the moment I am not going to have the time to attempt to re-do this off the current master. For now, I will close this and perhaps take another crack at it at a later date.

Regardless, thank you for the excellent library/framework - it's been of great benefit for generating documentation in a maintainable way!

Issue #506, opened 2018-04-28
Sample project not packaging static html in jar


thomastodon:
I'm not seeing the generated restdocs getting packaged within the jar. Is there something I'm doing wrong here?

Encountered issue at `a1198069beffdceec0b492b497168dd40fe6c885`

**Steps to Reproduce:**
1. navigate to the `junit5` example
2. `$ ./gradlew clean build`
3. `$ jar -tvf build/libs/junit5.jar`

**Expected Result:**
```
...
     0 Sat Apr 28 13:04:44 CDT 2018 META-INF/
   139 Sat Apr 28 13:04:44 CDT 2018 META-INF/MANIFEST.MF
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/example/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/example/junit5/
   928 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/example/junit5/SampleJUnit5Application.class
   762 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/example/junit5/SampleJUnit5Application$SampleController.class
     0 Sat Apr 28 13:15:16 CDT 2018 BOOT-INF/classes/static/
     0 Sat Apr 28 13:15:16 CDT 2018 BOOT-INF/classes/static/docs/
 32064 Sat Apr 28 13:15:16 CDT 2018 BOOT-INF/classes/static/docs/index.html
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/lib/
...
```

**Actual Result:**
```
...
     0 Sat Apr 28 13:04:44 CDT 2018 META-INF/
   139 Sat Apr 28 13:04:44 CDT 2018 META-INF/MANIFEST.MF
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/
     0 Sat Apr 28 13:04:44 CDT 2018 BOOT-INF/classes/com/
[...]

PR #518, opened 2018-06-19
 task name jar does not match boot 2.0

jojoldu:
The jar Task in build.grade is from 2.0 to bootJar.

Changed to jar -> bootJar.


wilkinsona:
Thanks for opening your first Spring REST Docs pull request. There's a bit more to this one as the samples are affected too (see #506). Could you update your pull request to correct the samples too please?

jojoldu:
Thanks for the reply!
As you said, I modified the sample project build.gradle as well.
I did not see the revision history on the PR page, so I created a [new PR](https://github.com/spring-projects/spring-restdocs/pull/521).
If not, I will delete it.

PR #521, opened 2018-06-26
Jojoldu/jar task name fix

jojoldu:
The jar Task in build.grade is from 2.0 to bootJar.

Changed to jar -> bootJar


jojoldu:
@wilkinsona Thanks!

PR #526, opened 2018-07-03
Remove static modifier for enums

izeye:
This PR removes `static` modifier for `enum`s.


wilkinsona:
Thank you once again, @izeye. This change is now in both 1.2.x and master.

PR #530, opened 2018-07-20
Use ID-based equality check for TemplateFormat consistently

izeye:
There are three ways to check equality for `TemplateFormats` constants:

- `getId().equals()`
- `equals()`
- `==`

As they are constants, `==` will be sufficient. So this PR changes to use `==` for equality check consistently.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/templates/TemplateFormats.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/templates/TemplateFormats.java
@@ -17,7 +17,7 @@
 package org.springframework.restdocs.templates;
 
 /**
- * An enumeration of the built-in formats for which templates are provuded.
+ * An enumeration of the built-in formats for which templates are provided.
  *
  * @author Andy Wilkinson
  * @since 1.1.0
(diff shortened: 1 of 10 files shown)


wilkinsona:
Thanks, @izeye. While their are constants for Asciidoctor and Markdown, there's nothing to stop someone creating their own `TemplateFormat` implementation with the same ID as one of those constants. If someone does that, their `TemplateFormat` should be considered equal to the built-in one. In other words, `getId().equals()` is the right comparison to use in main code. It's less clear-cut in the tests as then we know we'll only be dealing with the constants.

izeye:
@wilkinsona Thanks for the feedback! If so, how about implementing `TemplateFormat.equals()` with `TemplateFormat.getId()` and using `TemplateFormat.equals()` consistently?

wilkinsona:
With `TemplateFormat` being an interface, it's difficult to enforce an implementation of `equals`. I think I'd prefer to keep the existing id-based equality check in the main code. I don't really feel strongly about the tests. Given that the constants exist and they're expected to be used, a same instance check is not unreasonable. On the other hand, lots of the tests shouldn't really know or care about the constants and a id-based equality check would keep things more loosely coupled.

wilkinsona:
Sorry, I wasn't very clear. I think the test's current use of `equalTo` to compare `TemplateFormat` instances could be improved. It should either use `sameInstance` (thereby verifying that the constant is being used), or it should compare the formats' IDs.

I think I prefer the latter as it minimises any coupling to the constants.

izeye:
@wilkinsona Thanks for the clarification! I updated this as you suggested.

izeye:
I missed two places and polished them a bit in 4e75e0093ba6348cf5294763e315449b7014603c.

wilkinsona:
Thanks again, @izeye. These changes are now in 1.2.x and master.

PR #535, opened 2018-07-30
Polish JsonContentHandler.isEmptyCollection()

izeye:
This PR polishes `JsonContentHandler.isEmptyCollection()` by removing a removable flow.


wilkinsona:
Thanks, @izeye. This was intentional when I wrote it as I thought it made the intent clearer. Your proposal's made me doubt myself now though. I'm not sure the current code is any clearer, and it's more verbose.

izeye:
> I'm not sure the current code is any clearer and it's more verbose.

@wilkinsona When I saw the code, I felt the same thing but on the other hand also felt like a duplicate of the following flow. I don't have a strong opinion on this and am okay with either way.

wilkinsona:
Thanks again, @izeye. The proposed change is now in 1.2.x and master.

PR #543, opened 2018-08-15
Remove unused gradleWrapperVersion property from Grails sample

izeye:
This PR removes an unused `gradleWrapperVersion` property.


wilkinsona:
Thanks once again, @izeye.

PR #549, opened 2018-09-05
Add public FieldTypeResolver.

mduesterhoeft:
This can be useful for extensions to reuse field type determination logic.


mduesterhoeft:
@wilkinsona this is what we talked about in the chat the other day. Is that going into the right direction. 

🤔 The Codacy complaint is not valid I think - and I just moved the code around. But if you want me to change the enum comparison from `==` to `equals` I would be happy to do this.

wilkinsona:
@mduesterhoeft Thanks for the PR and sorry for the annoyance with Codacy. It's been irritating me for a little while now. Your comment has given me the nudge that I needed to remove the integration.

mduesterhoeft:
@wilkinsona no worries - I always struggle with myself if such code analysis is useful or not. I guess sometimes it is - sometimes it is rather annoying.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/ContentTypeHandlerFactory.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2014-2016 the original author or authors.
```
wilkinsona: This should be 2014-2018.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldTypeResolver.java
```diff
@@ -0,0 +1,48 @@
...
+package org.springframework.restdocs.payload;
+
+import org.springframework.http.MediaType;
+
+/**
+ * Public abstraction for external access to field type determination for xml and json
+ * payloads.
+ *
+ * @author Mathias Düsterhöft
+ */
```
wilkinsona: This is new public API so it should be annotated with `@since`. I think I'd like to sneak this into 2.0.x, so it should be `@since 2.0.3`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldTypeResolver.java
```diff
@@ -0,0 +1,48 @@
...
+public interface FieldTypeResolver {
+
+	/**
+	 * Create a FieldTypeResolver for the given content and contentType.
+	 * @param content the payload that the {@link FieldTypeResolver} should handle
+	 * @param contentType the content type of the payload
+	 * @return the {@link FieldTypeResolver}
+	 */
+	static FieldTypeResolver create(byte[] content, MediaType contentType) {
```
wilkinsona: I think a name like `forContent` may read a little better. What do you think?

spring-restdocs-core/src/test/java/org/springframework/restdocs/payload/FieldTypeResolverTests.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Tests for {@link FieldTypeResolver}.
+ *
+ * @author Mathias Düsterhöft
+ */
+public class FieldTypeResolverTests {
+
+	@Test
+	public void returnJsonFieldTypeResolver() {
+		assertThat(FieldTypeResolver.create("{\"field\": \"value\"}".getBytes(),
```
wilkinsona: Could you use `ExpectedException` here please? I'd like to move to AssertJ's `assertThatExceptionOfType`, but until that change is made across the whole codebase I'd prefer to make consistent use of `ExceptedException`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/ContentTypeHandlerFactory.java
```diff
@@ -0,0 +1,67 @@
...
+package org.springframework.restdocs.payload;
+
+import org.springframework.http.MediaType;
+
+/**
+ * Factory providing access to {@link ContentHandler} implementation for the content type.
+ *
+ * @author Mathias Düsterhöft
+ */
+final class ContentTypeHandlerFactory {
```
wilkinsona: I think this could be replaced with a static `forContent` method on `ContentHandler`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldTypeResolver.java
```diff
@@ -0,0 +1,49 @@
...
+				.getFieldTypeResolver();
+	}
+
+	/**
+	 * Returns the type of the field that is described by the given
+	 * {@code fieldDescriptor} based on the content of the payload.
+	 * @param fieldDescriptor the field descriptor
+	 * @return the type of the field
+	 */
+	Object determineFieldType(FieldDescriptor fieldDescriptor);
```
wilkinsona: For consistency with the other `Resolver` classes in REST Docs, this method should be named `resolveFieldType`.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/JsonContentHandler.java
```diff
@@ -144,30 +147,4 @@ private boolean isEmpty(Object object) {
 		return ((List<?>) object).isEmpty();
 	}
 
-	@Override
-	public Object determineFieldType(FieldDescriptor fieldDescriptor) {
```
wilkinsona: I think this logic should remain here. It's subtle, but there's a difference between resolving the type of a field (which is what the field type resolver does) and determining the type of a field. It's only the latter that considers the type that's specified on the descriptor and checks whether or not they match. Ideally the former would only use the path (it currently looks at whether the field is optional too). I'd like to keep the current level of separation.

spring-restdocs-core/src/test/java/org/springframework/restdocs/payload/JsonContentHandlerTests.java
```diff
@@ -29,75 +29,13 @@
...
  * @author Andy Wilkinson
+ * @author Mathias Düsterhöft
  */
 public class JsonContentHandlerTests {
 
 	@Rule
 	public ExpectedException thrown = ExpectedException.none();
 
-	@Test
-	public void typeForFieldWithNullValueMustMatch() {
```
wilkinsona: Given my comment above, I think these tests should remain here.

spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/ContentTypeHandlerFactory.java
mduesterhoeft: Are you sure? - `ContentHandler` is an interface at the moment. So we could use a default method. Then the factory method feels a little odd when working with the implementations. 

That is why I introduced the factory class. 

We could also make `ContentHandler` an abstract class.

Let me know which way you prefer.

mduesterhoeft:
@wilkinsona I refactored to be able to keep the separation of concerns of `JsonContentHandler` and `JsonFieldTypeResolver`. I now try to just use the existing code as much as possible. What do you think? 

I have the feeling that now `FieldTypeResolver` is not an ideal name, because compared to `JsonFieldTypeResolver` it is doing something different. Should we still keep the name?

wilkinsona:
Thanks for your patience, @mduesterhoeft.

> I have the feeling that now FieldTypeResolver is not an ideal name, because compared to JsonFieldTypeResolver it is doing something different. Should we still keep the name?

I agree that it's not ideal. having a `FieldTypeResolver` interface that isn't implemented by `JsonFieldTypeResolver` does not feel right to me. What's actually exposed is the field type determination logic from the content handlers, rather than the type resolution logic from `JsonFieldTypeResolver`. As [mentioned earlier](
https://github.com/spring-projects/spring-restdocs/pull/549#discussion_r222643345), there's a subtle yet important difference between the two and I think it's important that different names are used to help to maintain that distinction.

Looking back at the discussion in Gitter, it's the logic that's currently in the content handler that you're interested in, so what's exposed is what you want. However, the internal structure isn't quite right. I'll need to take another look at this to see how best it can be achieved. It may be that we need to take a different approach.

wilkinsona:
Thanks very much for your contribution, @mduesterhoeft. In the end I [reworked `JsonFieldTypeResolver`](
https://github.com/spring-projects/spring-restdocs/commit/ada273ba9eadb00705263fba0489086a965076b1) to be something that just does discovery and renamed it accordingly. This meant that your proposal could drop on top of that pretty much unchanged. All the changes are now in master and will be in the 2.0.3 release.

PR #550, opened 2018-09-05
Cache gradle dir to speed up travis builds

mduesterhoeft:


wilkinsona:
Thanks. The proposed changes are now in 1.2.x and master.

PR #553, opened 2018-10-03
Correct typo in default output directory

VonUniGE:
Default output directory is generated-snippets and not generated-snippet.


wilkinsona:
Well spotted. Thank you very much for making your first contribution to Spring REST Docs.

PR #577, opened 2019-02-07
Modernizing the documentation.

Buzzardo:
Now it pulls from spring-doc-resources to get the stylesheet, JavaScript, and static HTML that the modern look and feel uses, and it applies those items to the reference guide.


Buzzardo:
Superceded by https://github.com/spring-projects/spring-restdocs/pull/671

PR #589, opened 2019-03-01
Add sample for using WebFlux's WebTestClient with JUnit5

geminiKim:
I am using `WebFlux with JUnit5` and I want to use `spring-restdocs` and did not have an example for this, so I added sample for using WebFlux's WebTestClient with JUnit5.

If you could review this PR then I appreciate it, `spring-restdosc` is very cooool project. 
Thank you for your effort. :D


wilkinsona:
Thanks for the PR and for the kind words about REST Docs. To keep the numbers of samples manageable, I don't want to have one for every combination. There are already separate samples for JUnit 5 and WebTestClient so I think what we've got now is sufficient. Thanks anyway.

geminiKim:
@wikinsona I understand enough, thanks for the quick feedback.

geminiKim:
@wilkinsona I understand enough, thanks for the quick feedback.

Issue #581, opened 2019-02-15
spring-restdocs-asciidoctor is incompatible with AsciidoctorJ 1.6 and later


jnizet:
Using the latest version (1.5.10) of the gradle plugin `org.asciidoctor.convert` causes the following exception to be thrown:
```
Exception in thread "main" java.lang.IncompatibleClassChangeError: Found interface org.asciidoctor.extension.JavaExtensionRegistry, but class was expected
        at org.springframework.restdocs.asciidoctor.RestDocsExtensionRegistry.register(RestDocsExtensionRegistry.java:32)
        at org.asciidoctor.extension.internal.ExtensionRegistryExecutor.registerAllExtensions(ExtensionRegistryExecutor.java:21)
        at org.asciidoctor.internal.JRubyAsciidoctor.registerExtensions(JRubyAsciidoctor.java:113)
        at org.asciidoctor.internal.JRubyAsciidoctor.processRegistrations(JRubyAsciidoctor.java:102)
        at org.asciidoctor.internal.JRubyAsciidoctor.create(JRubyAsciidoctor.java:82)
        at org.asciidoctor.internal.JRubyAsciidoctor.create(JRubyAsciidoctor.java:78)
        at org.asciidoctor.Asciidoctor$Factory.create(Asciidoctor.java:713)
        at org.asciidoctor.gradle.backported.AsciidoctorJavaExec.getAsciidoctorInstance(AsciidoctorJavaExec.groovy:71)
        at org.asciidoctor.gradle.backported.AsciidoctorJavaExec.run(AsciidoctorJavaExec.groovy:39)
        at org.asciidoctor.gradle.backported.AsciidoctorJavaExec.main(AsciidoctorJavaExec.groovy:194)
```

Using the previous version (1.5.9.2) works fine.

It would be a good idea I guess to release a new version of spring-restdocs that is built against the newest stable version of AsciidoctorJ (the gradle plugin currently uses AsciidoctorJ version 1.6.1, and the latest version is 1.6.2). Unfortunately, the Maven plugin still uses AsciidoctorJ version 1.5.8. Not sure if it's possible to be compatible with the two versions.

wilkinsona:
[...]

PR #628, opened 2019-08-02
Section title out of sequence warnings may be generated when using operation macro

ahus1:
I've run the sample rest-notes-spring-hateoas locally using `gradlew asciidoctor` and saw that it generated lots of error messages like

```
asciidoctor: WARNING: api-guide.adoc: line 2: section title out of sequence: expected levels 0 or 1, got level 4
```

I traced this down to `operation_block_macro.rb` and changed it so that the block level is adjusted after `Asciidoctor.load` when the blocks are added to the parent. I assume this is safe when there are no nested blocks. 

Once I change it, the output is still correct and the warning message is gone.

### DIFF
--- a/spring-restdocs-asciidoctor-support/src/main/resources/extensions/operation_block_macro.rb
+++ b/spring-restdocs-asciidoctor-support/src/main/resources/extensions/operation_block_macro.rb
@@ -37,10 +37,9 @@ class OperationBlockMacro < Asciidoctor::Extensions::BlockMacroProcessor
 
   def do_read_snippets(snippets, parent, operation, snippet_titles)
     content = StringIO.new
-    section_level = parent.level + 1
     section_id = parent.id
     snippets.each do |snippet|
-      append_snippet_block(content, snippet, section_level, section_id,
+      append_snippet_block(content, snippet, section_id,
                            operation, snippet_titles)
     end
     content.string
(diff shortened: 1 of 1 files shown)


spring-restdocs-asciidoctor-support/src/main/resources/extensions/operation_block_macro.rb
```diff
@@ -96,8 +96,8 @@ def write_content(content, snippet, operation)
     end
   end
 
-  def write_title(content, snippet, level, id, snippet_titles)
-    section_level = '=' * (level + 1)
+  def write_title(content, snippet, id, snippet_titles)
+    section_level = '=='
```
wilkinsona: Switch to hardcoding to two `=` feels like it may cause a regression to me. For example, will it not now be wrong if the macro is used in a deeply nested section?

ahus1: The new line 53 will adjust the level of all top level blocks the moment they are inserted. This avoid the warning that would otherwise occur when starting with a section of a too high level, and avoid having a regresseion.

luvarqpp:
I am using this patch and it is working for me. I would be happy to see it in milestone [2.0.4.RELEASE](https://github.com/spring-projects/spring-restdocs/milestone/28) along with https://github.com/spring-projects/spring-restdocs/issues/581 solved.

wilkinsona:
Thanks for the PR, @ahus1. The proposed changes are now in master along with a [polishing commit](https://github.com/spring-projects/spring-restdocs/commit/9de06a36d225ec8aedf68641384021f278cbbb29) that updated the tests to cover the scenario that previously generated a warning.

PR #629, opened 2019-08-02
Fix IDs of sections in examples

ahus1:
In the examples the sections had their IDs in single brackets. 

This way they would be block attributes or block types. 

To set the ID for section, double brackets are necessary.

This PR fixes it.

(I found this as Asciidoctor 2.x versions complain about the block type)


wilkinsona:
Good catch. Thanks very much for making your first contribution to Spring REST Docs.

PR #638, opened 2019-09-09
Section number incremented when it shouldn't

ahus1:
My pull request for #628 introduced a regression: when using numbered sections using :sectnums: this produced gaps in the chapter number. 

I traced it down to the following issue: when the block is created, it is assigned a "sectname". For "==" this is "chapter", and it will increase the chapter number.

This change now sets level and sectname from a temporary template node. This leads to the correct sectname, and no chapter number will be increased. 

I see that #628 has been picked for the next maintenance release; maybe you want to revert/postpone it in the branch until this is taken care of.

I marked this [WIP] as it is missing a test.


ahus1:
I've updated an existing test case to cover also the chapter numbering. As this has now a test, I removed the WIP.

PR #651, opened 2019-10-16
Remove accidental characters in sample's pom.xml

izeye:
This PR removes accidental characters in sample's `pom.xml` as it seems to be introduced accidentally.

### DIFF
--- a/samples/rest-notes-spring-data-rest/pom.xml
+++ b/samples/rest-notes-spring-data-rest/pom.xml
@@ -86,7 +86,7 @@
 							<logHandler>
 								<outputToConsole>true</outputToConsole>
 								<failIf>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks very much, @izeye. I'm rather amazed that the sample still built with that extra `(2)` in it.

izeye:
@wilkinsona My pleasure 😄 By the way did you find how the builds have passed with them before? I'm curious as I just imagined that the sample hasn't been included in the builds somehow when I was opening this PR in my iPad.

wilkinsona:
I haven't figured it out, but I did confirm that the sample builds fine locally with the `(2)` in there. It's also fine on CI where the samples are built as part of every build. I guess Maven ignores it somehow 🤷‍♂.

Issue #263, opened 2016-06-17
cannot use spring-restdocs with GsonHttpMessageConverter


BK-Choi:
My spring project works fine,
but tests crashes when I add spring-restdocs to gradle dependencies.
Just that (no configuration chages, no test code changes) brakes my project.

The problem is that spring-restdocs library make spring framework ignore `mvc:message-converters` configuration. (I'm not sure what is exactly happened)
I added gson over jackson for JSON as you can see below.
but with spring-restdocs, only I see is `messageConverters` without `GsonHttpMessageConverter`.

I have configuration like this:
```xml
<beans xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:context="http://www.springframework.org/schema/context"
       xmlns:mvc="http://www.springframework.org/schema/mvc"
       xsi:schemaLocation="
        http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc-4.2.xsd
        http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
        http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-4.2.xsd">

<!-- message converter -->
	<mvc:annotation-driven content-negotiation-manager="contentNegotiationManager">
		<mvc:message-converters>
			<ref bean="jaxbMessageConverter" />
			<ref bean="gsonMessageConverter" />
		</mvc:message-converters>
	</mvc:annotation-driven>

	<bean id="gsonMessageConverter" class="org.springframework.http.converter.json.GsonHttpMessageConverter"/>
	<bean id="jaxbMessageConverter" class="org.springframework.http.converter.xml.MarshallingHttpMessageConverter">
		<property name="marshaller" ref="jaxbMarshaller" />
		<property name="unmarshaller" ref="jaxbUnmarshaller" />
		<property name="supportedMediaTypes">
			<list>
				<value>application/xml</value>
				<value>application/xml;charset=UTF-8</value>
			</list>
		</property>
	</bean>

	<bean id="jaxbMarshaller" class="org.springframework.oxm.jaxb.Jaxb2Marshaller">
		<property name="packagesToScan">
			<list>
				<value>foo.boo.**.output</value>
			</list>
		</property>
	</bean>
[...]

PR #667, opened 2020-01-21
Allow to easily configure custom ObjectMapper to preprocessors

ninja-inc:
I would like to configure custom `ObjectMapper` to preprocessors easily because `PrettyPrintingContentModifier` supports only indent output now.

```
private WebTestClient webTestClient;
private ObjectMapper objectMapper;

@Before
public void setup() {
	this.webTestClient = WebTestClient.bindToApplicationContext(this.context)
		.configureClient()
		.filter(documentationConfiguration(this.restDocumentation)
			.operationPreprocessors()
				.withRequestDefaults(removeHeaders("Foo")) 
				.withResponseDefaults(customJsonPrint(objectMapper)))   // here!!
		.build();
}
```

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/operation/preprocess/CustomPrintingJsonContentModifier.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/operation/preprocess/CustomPrintingJsonContentModifier.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2014-2020 the original author or authors.
+ *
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the suggestion. Unfortunately, adding a central public API that depends on a Jackson type will further complicate #263 so I don't think we should do it. If you'd like to explore alternative ways of achieving your goal, please feel free to open an issue and we can discuss it. I feel that you may just be better having your own `ContentModifier` implementation that meets your needs.

PR #699, opened 2020-10-03
Warnings logged by the operation block macro lack context

ahus1:
I'm backporting a change that I applied in the IntelliJ AsciiDoc plugin to highlight problems with the operation macro in the editor. 

When using `logger.warn` with a cursor the output contains the file and line number where the problem occurred. 

The unit tests will log an example output like this (from a unit test)

```
WARNING: <stdin>: line 1: Snippet missing-snippet not found at C:\Users\...\AppData\Local\Temp\junit8239829976219138844\gradle-project\build\generated-snippets/some-operation/missing-snippet.adoc for operation some-operation
```


wilkinsona:
Thanks very much, @ahus1.

PR #700, opened 2020-10-03
"included in the package" cross-reference is broken

ahus1:
The anchor `getting-started-build-configuration-maven-packaging` is broken as it doesn't point an ID defined in one of the documents.
I tried to replace it from the context with `getting-started-build-configuration-packaging-the-documentation`. 

Please double-check if this is correct.


wilkinsona:
Thanks again, @ahus1.

PR #714, opened 2021-04-06
Fix bug when using beneath path and wildcard together

ddaaac:
Hi, I'm using restdocs so usefully 😀

I encountered following situation.

Here is the json body used in the request.
```json
{
   "number":"123",
   "link":{
      "scheme":"scheme",
      "url":"url"
   },
   "chat":{
      "message":"send message",
      "id":"id",
      "thread":{
         "content":"thread content",
         "count":1
      }
   }
}
```
And here is the test code using beneathPath and wildcard.
```kotlin
val messageRequest = MessageRequest(
        number = "123",
        link = Link("scheme", "url"),
        chat = Chat("send message", "id", Thread("thread content", 1)),
)
mockMvc.perform(
        RestDocumentationRequestBuilders.get("/test")
                .contentType(MediaType.APPLICATION_JSON)
                .content(objectMapper.writeValueAsBytes(messageRequest))
)
        .andExpect(status().isOk)
        .andDo(MockMvcRestDocumentation.document(
                "test",
                requestFields(
                        fieldWithPath("number").type(JsonFieldType.STRING).description("a"),
                        subsectionWithPath("link").type(JsonFieldType.OBJECT).description("z"),
                        subsectionWithPath("chat").type(JsonFieldType.OBJECT).description("f"),
                ),
                requestFields(
                        beneathPath("*.thread").withSubsectionId("thread"),
                        fieldWithPath("content").type(JsonFieldType.STRING).description("thread message").optional(),
                        fieldWithPath("count").type(JsonFieldType.NUMBER).description("count").optional(),
                ),
        ))
```
And I saw error like below.
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class java.lang.Object and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS)
```
This is because using wildcards will result in unmatched-pattern object field `link` being `Extracted Field.Absent`. `Extracted Field.Absent` is assigned to `Object()`, so serialization fail.
So I filter non-absent field.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldPathPayloadSubsectionExtractor.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/payload/FieldPathPayloadSubsectionExtractor.java
@@ -90,7 +90,7 @@ public class FieldPathPayloadSubsectionExtractor
 							(descriptor) -> JsonFieldPath.compile(this.fieldPath + "." + descriptor.getPath()),
 							this::prependFieldPath));
 			if (value instanceof List) {
-				List<?> extractedList = (List<?>) value;
+				List<?> extractedList = getNotAbsentExtractedField(value);
 				JsonContentHandler contentHandler = new JsonContentHandler(payload, descriptorsByPath.values());
 				Set<JsonFieldPath> uncommonPaths = JsonFieldPaths.from(extractedList).getUncommon().stream()
 						.map((path) -> JsonFieldPath.compile(this.fieldPath + "." + path)).filter((path) -> {
@@ -114,6 +114,17 @@ public class FieldPathPayloadSubsectionExtractor
 		catch (IOException ex) {
 			throw new PayloadHandlingException(ex);
 		}
+	}
+
+	private List<?> getNotAbsentExtractedField(Object value) {
+		List<?> extractedList = ((List<?>) value)
+				.stream()
+				.filter(field -> !ExtractedField.ABSENT.equals(field))
+				.collect(Collectors.toList());
+		if (extractedList.isEmpty()) {
+			throw new PayloadHandlingException(this.fieldPath + " does not identify a section of the payload");
+		}
+		return extractedList;
 	}
 
 	private FieldDescriptor prependFieldPath(FieldDescriptor original) {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the proposal. Purely based on the example above, the usage of `*` doesn't feel quite right to me. It's intended for situations where there's a common structure beneath one or more unknown keys. In this case, the structure doesn't appear to be common with `chat` being the only entry that contains `thread`. It's this lack of common structure that's causing the failure. The error message could certainly be improved, but I'm not sure that the functionality should be changed.

Let's take a bit of a step back. What led to you using `*.thread` rather than `chat.thread` when documenting the `thread` subsection?

ddaaac:
It is the minimum code to reproduce.
```json
{
   "number":"123",
   "link":{
      "scheme":"scheme",
      "url":"url"
   },
   "chat":{
      "message":"send message",
      "id":"id",
      "thread":{
         "content":"thread content",
         "count":1
      }
   },
     "chat2":{
      "message2":"send message",
      "thread":{
         "content":"thread content2",
         "count":2
      }
   },
   ...
}
```
There are several fields with the same structure as `*.thread`, and I wanted to document them at once. Is there a better way? 

----
And even though this is intended, it would be nice if a user-friendly error message was displayed. Currently, the user should know that ABSENT is Object().

wilkinsona:
> There are several fields with the same structure as *.thread, and I wanted to document them at once. Is there a better way?

Generally speaking, when something is difficult to document, that's also an indication that it will be difficult for clients to consume. If you have control over the format of the response, I would recommend grouping all of the `chat…` entries together. Something like this:

```
{
  "number":"123",
  "link":{
     "scheme":"scheme",
     "url":"url"
  },
  "chats":{
    "chat":{
       "message":"send message",
       "id":"id",
       "thread":{
          "content":"thread content",
          "count":1
       }
    },
    "chat2":{
      "message":"send message",
      "thread":{
         "content":"thread content2",
         "count":2
       }
    },
  }
}
```

You can then use `chats.*.thread`, `chats.*.message`, etc to document the structure of each chat. The more consistent structure will also make it easier for clients to map the response to objects, particularly in a type-safe language.

If changing the structure of the response isn't possible, you could document `number`, and `link` and then note that everything else should be treated as a chat. You could use [`relaxedResponeFields`](https://docs.spring.io/spring-restdocs/docs/2.0.5.RELEASE/api/org/springframework/restdocs/payload/PayloadDocumentation.html#relaxedResponseFields-org.springframework.restdocs.payload.FieldDescriptor...-) to document `number` and `link` and then perhaps `beneathPath` to focus on a chat and [document that subsection of the response](https://docs.spring.io/spring-restdocs/docs/2.0.5.RELEASE/reference/html5/#documenting-your-api-request-response-payloads-subsections).

> And even though this is intended, it would be nice if a user-friendly error message was displayed.

Yeah, I agree. I said above that "the error message could certainly be improved". I've opened https://github.com/spring-projects/spring-restdocs/issues/715 to do so.

PR #724, opened 2021-05-20
Remove redundant public modifiers for Groovy

izeye:
This PR removes redundant `public` modifiers for Groovy.

See https://groovy-lang.org/style-guide.html#_public_by_default


wilkinsona:
Thanks again, @izeye.

PR #766, opened 2021-12-08
Allow an empty body in response field when HTTP response code is NO_CONTENT

jervds:
In some API, it is possible to have for the same resource a response code OK (200) with a body or a response code NO_CONTENT (204) with an empty body.

When trying to document such resource for code 204, a failure occurs:
```
org.springframework.restdocs.snippet.SnippetException: Cannot document response fields as the response body is empty
```

This MR aims to allow to document such a resource with response code 204.


wilkinsona:
Thanks for the proposal but I am not sure that this fits with REST Docs' approach to generating documentation.

A fundamental goal of the approach is that your tests will fail if the service that you are documenting does not match the expectations in your tests. If merged, this change would mean that the goal would not be met if a response with no body is returned when the tests expect the response to have a body.

Rather than trying to have a single test that covers both a 200 response and a 204 response, I would recommend splitting it into two tests. One test would make a request that will result in a 200 response and document the response's fields. Another test would make a request that will result in a 204 response and assert that the body is empty.

Thanks anyway for the suggestion.

PR #819, opened 2022-06-30
Use HTTPS for REST Assured links

izeye:
This PR changes to use HTTPS for REST Assured links.

This PR also drops unnecessary `nohttp` `allowlist.lines` file.


wilkinsona:
Thanks again, @izeye.

Issue #886, opened 2023-04-09
Replace factory with builder when creating operation requests/responses


AugustoRavazoli:
Both the `OperationRequestFactory` and `OperationResponseFactory` leads to unnecessary code in custom pre-processors:
```java
new OperationRequestFactory().create(
  uri,
  request.getMethod(),
  request.getContent(),
  headers,
  request.getParts(),
  cookies
);
```
A possible way to simplify the request creation is to use the Builder design pattern:
```java
new OperationRequestBuider(request)
  .withUri(uri)
  .withHeaders(headers)
  .withCookies(cookies)
  .build();
```
The `OperationRequestBuilder` will builds the new request using the request argument fields as defaults, like a `StringBuilder`.

wilkinsona:
While this is quite a low-level API that many people won't touch, I can see the value in making it more concise and easier to use. Before doing so, I'd like the outcome of https://github.com/spring-projects/spring-restdocs/issues/733 to be decided.

Issue #733, opened 2021-06-15
Use exchange rather than operation to refer to a request and response


wilkinsona:
When REST Docs began there was no term in the Spring portfolio describing a request being sent to an endpoint and a response being received so "operation" was used. Since then, "exchange" has been adopted in Spring Framework. I'd like REST Docs to switch to using "exchange", primarily because I think consistency across the two projects is important but also because I think exchange is a better term anyway.

To ease migration, I'd like to introduce the use of exchange in 2.1 and deprecate the use of operation. I'm not sure how complex this will prove to be. There's quite a bit of code in the `org.springframework.restdocs.operation` and `org.springframework.restdocs.operation.preprocess` packages although not all of it is public or commonly used. There's also the operation Asciidoctor macro to consider.

PR #889, opened 2023-04-29
Add builders as alternatives to factories when creating operation req…

AugustoRavazoli:
…uests and responses

The intent of this change is to provide a better api to work with, dispensing repeated code from the factories

Fixes gh-886


wilkinsona:
Thanks for the proposal but I'd like the outcome of https://github.com/spring-projects/spring-restdocs/issues/733 to be decided before adding any new API in this area.

PR #916, opened 2024-01-23
Remove unused String.format call

mnhock:


wilkinsona:
Thanks again, @mnhock.

PR #921, opened 2024-03-05
fix: add form parameter content when the contentType is multipart/for…

### LINKED ISSUE #922: Cannot document form parameters when contentType is multipart/form-data
Hello,

In the case of requests in multipart form format, the form parameter is not documented.
An example of the request is as follows.

fix in #921 

```
Map<String, Object> params = new HashMap<>();
		params.put("password", "oldPassword");
		params.put("dvcid", "testDeviceId");

		// when
		MockMvcResponse response = given()
			.contentType(MediaType.MULTIPART_FORM_DATA_VALUE)
			.multiPart("profileImage", "test photo", "image/png")
			.formParams(params)
			.when()
			.post("/test/modify");

		response.prettyPrint();

		response.then()
			.apply(
				document("test-modify",
					ApiDocumentUtils.getDocumentRequest(),
					ApiDocumentUtils.getDocumentResponse(),
					formParameters(
						parameterWithName("oldPassword").attributes(emptyFormat()).description("password"),
						parameterWithName("dvcid").attributes(emptyFormat()).description("dvcid")
					),
					requestParts(
						partWithName("profileImage").attributes(emptyFormat(), required()).description("프로필 이미지 파일")
					)
				)
			)
			.statusCode(HttpStatus.OK.value());
```

fucct:
fix #922

### DIFF
--- a/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcRequestConverter.java
+++ b/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcRequestConverter.java
@@ -16,40 +16,27 @@
 
 package org.springframework.restdocs.mockmvc;
 
-import java.io.IOException;
-import java.net.URI;
-import java.net.URLDecoder;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Scanner;
-
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.Part;
-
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR. There appear to be two different GitHub users here, `fucctt` and `fucct`. Can you please make sure that the PR is opened by the same user who authored the commits that it contains?

fucct:
oh, I just fix it.

wilkinsona:
Thanks for the PR and for your patience while I found time to be able to look at it.

Unfortunately, I don't think the fix that's proposed here is the right one. Looking at our example in #922, calling both `multiPart("profileImage", "test photo", "image/png")` and `formParams(params)` should result in a `multipart/form-data` request with three parts named `profileImage`, `password` and `dvcid`. The tests that you're proposing here are asserting that the content type will be `multipart/form-data` but with a body that's actually `application/x-www-form-urlencoded`. I'll see #922 to explore an alternative fix.

PR #938, opened 2024-09-08
Update api-guide.adoc

sophearyrin-dev:
I want to add one more method to http verb which is put. because it's missing. And Below is my updated. Thank you!
image

### DIFF
--- a/samples/rest-notes-spring-data-rest/src/main/asciidoc/api-guide.adoc
+++ b/samples/rest-notes-spring-data-rest/src/main/asciidoc/api-guide.adoc
@@ -27,8 +27,11 @@ use of HTTP verbs.
 | `POST`
 | Used to create a new resource
 
+| `PUT`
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR. The 2.0.x branch in this repository is no longer maintained. The `main` branch does not contain any samples as they have moved to a [separate repository](https://github.com/spring-projects/spring-restdocs-samples). Would you like to open a similar PR over there? Before doing so, please be aware that I don't think it makes sense to add `PUT` only to the table of HTTP verbs as nowhere else in the documentation talks about `PUT` requests. To warrant its addition to the table of HTTP verbs, the documentation for updating notes and tags should also be updated to discuss `PUT` requests.

sophearyrin-dev:
Thank you @wilkinsona for your comment. May I know why PUT http method is not included in the document?

wilkinsona:
It's several years since it was written so I'm not certain, but I would guess that it was just an oversight when producing the Spring Data REST version of the sample.

Issue #887, opened 2023-04-15
How to document constraints based on groups validations?


AugustoRavazoli:
The docs gives an example of documenting constraints:
```java
public void example() {
	ConstraintDescriptions userConstraints = new ConstraintDescriptions(UserInput.class);
	List<String> descriptions = userConstraints.descriptionsForProperty("name");
}

static class UserInput {

	@NotNull
	@Size(min = 1)
	String name;

	@NotNull
	@Size(min = 8)
	String password;

}
```
This way, you get **all** constraints descriptions for one property, but when using validation groups, how to get only the constraints descriptions for the given group? For example:
```java
public void example() {
	ConstraintDescriptions userConstraints = new ConstraintDescriptions(UserInput.class);
        List<String> descriptions = userConstraints.descriptionsForProperty("email").groups(OnCreate.class);
}

static class UserInput {

	@NotNull
	String name;

        @NotNull(groups = OnCreate.class)
        @Null(groups = OnEdit.class)
        String email;
}
```

AugustoRavazoli:
After reading the source code of the library and some tests, I found a possible solution using a custom `ResourceBundleConstraintDescriptionResolver`:
```java
public class ConstrainedFields {

    private final ConstraintDescriptions constraints;
    private final List<Class<?>> groups;  // Constraints groups to exclude

    public ConstrainedFields(Class<?> clazz) {
      groups = new ArrayList<>();
      constraints = new ConstraintDescriptions(clazz, resolver(groups));
    }

    public FieldDescriptor pathExcludingGroups(String path, Class<?>... excludedGroups) {
      groups.addAll(List.of(excludedGroups));
      return path(path);
    }

    public FieldDescriptor path(String path) {
      var description = constraints.descriptionsForProperty(path).stream()
        .filter(s -> !s.isEmpty())
        .collect(joining(". "));
      return fieldWithPath(path).attributes(key("constraints").value(description));
    }

    private static ResourceBundleConstraintDescriptionResolver resolver(List<Class<?>> groups) {
      return new ResourceBundleConstraintDescriptionResolver() {

        @Override
        public String resolveDescription(Constraint constraint) {
          var description = super.resolveDescription(constraint);
          if (!groups.isEmpty()) {
            var constraintGroups = (Class<?>[]) constraint.getConfiguration().get("groups");
            if (List.of(constraintGroups).containsAll(groups)) {
              description = "";
              groups.clear();
            }
          }
          return description;
        }

      };
    }

[...]

PR #940, opened 2024-09-17
Add support for Bean Validation's constraint groups

nosan:
gh-887

- Groups can be obtained from configuration (key: **groups**)
- As an alternative solution, `ConstraintDescriptions` can be extended with additional methods with `groups` as an argument. 

P.S. docs also need to be updated.

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/constraints/ValidatorConstraintResolver.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/constraints/ValidatorConstraintResolver.java
@@ -17,7 +17,9 @@
 package org.springframework.restdocs.constraints;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks, @nosan. Plenty of food for thought there. 

> REST Docs, however, currently returns all constraints. It's unclear whether this behavior is intentional.

It's too long ago need for me to remember the intent with any degree of certainty. Looking at the code and tests, I suspect that I just didn't consider groups at all and the current behaviour wasn't intended.

> I'm not sure that passing groups via the constructor is a good option here. While it works, it means that I need to create a new `ValidatorConstraintResolver` every time I want to use different groups, which, to be honest, is quite inconvenient.

I wonder if we should be making some more assumptions about how validation is triggered, certainly in the golden-path case. If we assume the use of Framework's `@Validated`, we could look for that on the class that's passed to `resolveForProperty` to determine the groups. This could also work for method parameter constraints, support for which [may be coming in the future](https://github.com/spring-projects/spring-restdocs/issues/1026).

Issue #942, opened 2024-10-01
Omit the version in the Maven Resources plugin example


dve:
In the documentation on "Packaging the Documentation" the introduction points to Spring Boot, but the code sample sets the version for the maven-resources-plugin which is managed by Spring Boot.

It would be nice if there is a comment that Spring Boot manages this version, so Spring Boot users should not declare it.

https://github.com/spring-projects/spring-restdocs/blob/d2461e737d4a54c4e8070fb64d0820d794c52066/docs/src/docs/asciidoc/getting-started.adoc?plain=1#L149

wilkinsona:
Thanks for the suggestion. I think it would be better to just omit it entirely as it's likely to be out of date (as it is right now).

PR #943, opened 2024-10-01
#942 Omit the version in the Maven Resources plugin example

dve:
Fix for #942


wilkinsona:
Thanks very much, @dve.

PR #948, opened 2024-10-23
Correct typo from 'Explict' to 'Explicit'

LeeSJ0109:
Correct typo from 'Explict' to 'Explicit'.


wilkinsona:
Thanks very much, @LeeSJ0109.

Issue #945, opened 2024-10-07
Add support for modifying a URI's path during pre-processing


wilkinsona:
See #944 for background. In addition to the existing support for modifying the URI's scheme, host, and port, it would also be useful to be able to manipulate its path to, for example, add a prefix.

big-cir:
Fixed PR: #952

PR #952, opened 2024-12-27
Add support for URI's path prefix modification

big-cir:
Overview
---
1. Added pathPrefix field and setter method to UriModifyingOperationPreprocessor
2. Added corresponding support in UriModifyingContentModifier

Related Issues
---
#945 

Test
---
```java
@Test
public void requestContentUriPathIsPreservedAndAddPathPrefix() {
    this.preprocessor
		    .removePort()
		    .pathPrefix("/addPathPrefix");
    OperationRequest processed = this.preprocessor
	    .preprocess(createRequestWithContent("The uri 'http://localhost:12345/foo/bar' should be used"));
    assertThat(new String(processed.getContent())).isEqualTo("The uri 'http://localhost/addPathPrefix/foo/bar' should be used");
}

@Test
public void responseContentUriPathIsPreservedAndAddPathPrefix() {
    this.preprocessor
		    .removePort()
		    .pathPrefix("addPathPrefix");
    OperationResponse processed = this.preprocessor
	    .preprocess(createResponseWithContent("The uri 'http://localhost:12345/foo/bar' should be used"));
    assertThat(new String(processed.getContent())).isEqualTo("The uri 'http://localhost/addPathPrefix/foo/bar' should be used");
}
```

### DIFF
--- a/spring-restdocs-core/src/main/java/org/springframework/restdocs/operation/preprocess/UriModifyingOperationPreprocessor.java
+++ b/spring-restdocs-core/src/main/java/org/springframework/restdocs/operation/preprocess/UriModifyingOperationPreprocessor.java
@@ -228,14 +228,15 @@ public class UriModifyingOperationPreprocessor implements OperationPreprocessor 
 		private String modify(String input) {
 			List<String> replacements = Arrays.asList(this.scheme, this.host,
 					StringUtils.hasText(this.port) ? ":" + this.port : this.port,
-					this.pathPrefix);
+					this.pathPrefix == null ? "" : this.pathPrefix.startsWith("/") ? this.pathPrefix : "/" + this.pathPrefix);
 
 			int previous = 0;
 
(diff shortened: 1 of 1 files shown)


big-cir:
Could you please let me know if pattern validation for prefixPath is also necessary in the modify method?

``` java
private static final Pattern SCHEME_HOST_PORT_PATTERN = Pattern
			.compile("(http[s]?)://([a-zA-Z0-9-\\.]+)(:[0-9]+)?");

private String modify(String input) {
   ...
   Matcher matcher = SCHEME_HOST_PORT_PATTERN.matcher(input);
}
```

wilkinsona:
Thanks for PR. Unfortunately, I'm not sure that focussing solely on the path's prefix is the right approach as there may be a need to modify the path in other ways. I think we need to take a step back and consider exactly what modifications to support before looking at possible implementations. One option may be to support something like Spring Framework's `UriComponentsBuilder`. Another would be to allow the user to provide a `UnaryOperator` that's called with the current path and can return a new path modified in any way they require.

Issue #953, opened 2025-01-06
 Duplicate File Parts Generated in MockMvc using MockPart 


Lee-WonJun:
MockMultipartHttpServletRequest and MockMultipartHttpServletRequestBuilder Behavior have been found with the behavior of `MockMultipartHttpServletRequest` (spring-test) and `MockMultipartHttpServletRequestBuilder` (spring-test). 

`MockMultipartHttpServletRequest` internally manages `multipartFiles` as separate file types. When using `MockMultipartHttpServletRequestBuilder` to create a mock multipart request, files can be added using either the `.part` method to add a `MockPart` or the `.file` method to add a `MockMultipartFile`.

```java
this.mockMvc.perform(multipart("/upload")
        .file(new MockMultipartFile("file", "file.pdf", "application/pdf", "sample content".getBytes()))
        .part(new MockPart("file", "test.pdf", "contents".getBytes(), MediaType.APPLICATION_PDF))
);
```

The actual servlet creation process is implemented as follows:

```java
protected final MockHttpServletRequest createServletRequest(ServletContext servletContext) {
    MockMultipartHttpServletRequest request = new MockMultipartHttpServletRequest(servletContext);
    Charset defaultCharset = (request.getCharacterEncoding() != null ?
            Charset.forName(request.getCharacterEncoding()) : StandardCharsets.UTF_8);

    this.files.forEach(request::addFile);
    this.parts.values().stream().flatMap(Collection::stream).forEach(part -> {
        request.addPart(part);
        try {
            String name = part.getName();
            String filename = part.getSubmittedFileName();
            InputStream is = part.getInputStream();
            if (filename != null) {
                request.addFile(new MockMultipartFile(name, filename, part.getContentType(), is));
            } else {
                InputStreamReader reader = new InputStreamReader(is, getCharsetOrDefault(part, defaultCharset));
                String value = FileCopyUtils.copyToString(reader);
                request.addParameter(part.getName(), value);
            }
        } catch (IOException ex) {
            throw new IllegalStateException("Failed to read content for part " + part.getName(), ex);
        }
    });
}
```

As can be seen in the code, files added using the `.file` method are only processed using `addFile`. However, files added using the `.part` method are processed using both `addPart` and `addFile`, resulting in duplicate data being included in the servlet creation.

### Spring Server and Multipart Data Processing

Spring servers can process multipart data in two ways:

1. `@RequestParam("file") MultipartFile file`
[...]

PR #954, opened 2025-01-06
Fix: Duplicate File Parts Generated in MockMvc using MockPart

Lee-WonJun:
PR for Issue #953

**Solution:**

1. Compare if the `File` and `Part` contain the same content. If they do, remove one of them (assuming they are paired, so once removed, it will not be compared again).
2. If they are not the same, add the `part` to the `parts` list.
3. The `MockPart` implementation itself calls `setContentDispositionFormData`, which is only added to `Part`. Therefore, when comparing, `setContentDispositionFormData` is also included in the comparison.

**Reference Code (MockPart):**

```java
public MockPart(String name, @Nullable String filename, @Nullable byte[] content, @Nullable MediaType contentType) {
		Assert.hasLength(name, "'name' must not be empty");
		this.name = name;
		this.filename = filename;
		this.content = (content != null ? content : new byte[0]);
		this.headers.setContentDispositionFormData(name, filename);
		this.headers.setContentType(contentType);
	}
```

**Additional Information:**

I initially wrote a test code to compare the `MediaType` content type, but due to compatibility checks, I could not use the `MockPart` constructor that allows defining `contentType` (introduced in `@since 6.1.2`). As a result, I removed it from the test.

### DIFF
--- a/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcRequestConverter.java
+++ b/spring-restdocs-mockmvc/src/main/java/org/springframework/restdocs/mockmvc/MockMvcRequestConverter.java
@@ -156,12 +156,45 @@ class MockMvcRequestConverter implements RequestConverter<MockHttpServletRequest
 
 	private List<OperationRequestPart> extractParts(MockHttpServletRequest servletRequest)
 			throws IOException, ServletException {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, but it's too early to consider possible fixes for #953. It's not yet clear if a change in REST Docs is even the right thing to do.

PR #962, opened 2025-05-12
Align bootJar example with changes to Asciidoctor's Gradle plugin

djarnis73:
I could not get /html5 to work but removing it made it work.

It is also what is done in this example https://github.com/spring-projects/spring-restdocs-samples/blob/7ccbf89908ce1726c9c7dd30d291bb61f95436a3/rest-assured/build.gradle#L49


wilkinsona:
Thanks very much, @djarnis73.

PR #970, opened 2025-06-20
Correct "Including Multiple Snippets for an Operation"

nicos68:
A couple of sentences in the documentation do not make sense. I reworded them the way I feel they should have been written.


wilkinsona:
Thanks very much, @nicos68.

Issue #1038, opened 2026-04-09
WebTestClientRequestConverter mishandles cookies whose value contains = and crashes on cookie headers without =


config25:
## Summary

  `WebTestClientRequestConverter.createRequestCookie(String)` parses the `Cookie` header value with a naive `split("=")` and unconditionally accesses `components[1]`. This leads to two related problems:

  1. **Data corruption** — when a cookie value legitimately contains `=` (e.g. Base64-padded session IDs, JWTs in some encodings), only the substring before the first `=` is captured as the value, and the remainder is silently
  dropped.
  2. **`ArrayIndexOutOfBoundsException`** — when a cookie header value contains no `=` at all, `components[1]` throws.

  ## Affected code

  `spring-restdocs-webtestclient/src/main/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverter.java`, lines 107–110:

  ```java
  private RequestCookie createRequestCookie(String header) {
      String[] components = header.split("=");
      return new RequestCookie(components[0], components[1]);
  }
  ```

  There is no length check on `components`, and `split("=")` is called without a limit, so any `=` inside the value is also treated as a delimiter.

  For comparison, `MockMvcRequestConverter.extractCookies(...)` reads cookies from the already-parsed `MockHttpServletRequest#getCookies()` and never touches the raw header string, so the equivalent path on the MockMvc side is
  unaffected.

  ## Scenario 1 — Cookie value containing `=` (data corruption)

  `=` is a valid character inside a cookie value per [RFC 6265 §4.1.1](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) (only `;`, `,`, whitespace, and a few control characters are excluded). It is also extremely
  common in practice: Base64 padding (`abc=`, `abcd==`) is the obvious example, and many session/JWT cookies end up containing it.

  This is reachable through the standard `WebTestClient.cookie(name, value)` API:

  ```java
  ExchangeResult result = WebTestClient.bindToRouterFunction(RouterFunctions.route(GET("/foo"), (req) -> null))
      .configureClient()
      .baseUrl("http://localhost")
      .build()
      .get()
      .uri("/foo")
      .cookie("sessionId", "YWJjZGVm==")   // Base64-padded value
      .exchange()
      .expectBody()
      .returnResult();

  OperationRequest request = this.converter.convert(result);

  // Expected: value == "YWJjZGVm=="
[...]

PR #1040, opened 2026-04-13
WebTestClientRequestConverter mishandles cookies whose value contains =

config25:
`WebTestClientRequestConverter.createRequestCookie()` uses `split("=")` to
  parse cookie headers. This causes two problems:

  - Cookie values containing `=` (e.g. Base64-padded values like `YWJjZGVm==`)
    are silently truncated at the first `=`
  - Cookie headers without `=` throw `ArrayIndexOutOfBoundsException`

  Replace `split("=")` with `indexOf('=')` to split on only the first `=` and
  handle the no-`=` case by returning an empty value.

  Fixes gh-1038

  base branch 3.0.x

### DIFF
--- a/spring-restdocs-webtestclient/src/main/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverter.java
+++ b/spring-restdocs-webtestclient/src/main/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverter.java
@@ -104,8 +104,8 @@ class WebTestClientRequestConverter implements RequestConverter<ExchangeResult> 
 	}
 
 	private RequestCookie createRequestCookie(String header) {
-		String[] components = header.split("=");
-		return new RequestCookie(components[0], components[1]);
+		int separator = header.indexOf('=');
+		return new RequestCookie(header.substring(0, separator), header.substring(separator + 1));
 	}
 
 	private final class ExchangeResultReactiveHttpInputMessage implements ReactiveHttpInputMessage {
--- a/spring-restdocs-webtestclient/src/test/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverterTests.java
+++ b/spring-restdocs-webtestclient/src/test/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverterTests.java
@@ -335,4 +335,22 @@ public class WebTestClientRequestConverterTests {
 		assertThat(request.getCookies()).extracting("value").containsExactly("cookieVal1", "cookieVal2");
 	}
 
+	@Test
+	public void requestWithCookieValueContainingEqualsSign() {
+		ExchangeResult result = WebTestClient.bindToRouterFunction(RouterFunctions.route(GET("/foo"), (req) -> null))
+			.configureClient()
+			.baseUrl("http://localhost")
+			.build()
+			.get()
+			.uri("/foo")
+			.cookie("sessionId", "YWJjZGVm==")
+			.exchange()
+			.expectBody()
+			.returnResult();
+		OperationRequest request = this.converter.convert(result);
+		assertThat(request.getCookies()).hasSize(1);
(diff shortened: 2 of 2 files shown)


spring-restdocs-webtestclient/src/test/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverterTests.java
```diff
@@ -335,4 +335,40 @@ public void requestWithCookies() {
 		assertThat(request.getCookies()).extracting("value").containsExactly("cookieVal1", "cookieVal2");
 	}
 
+	@Test
+	void requestWithCookieValueContainingEqualsSign() {
```
wilkinsona: This method has to be `public` as the 3.0.x branch uses JUnit 4.

spring-restdocs-webtestclient/src/test/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverterTests.java
```diff
@@ -335,4 +335,40 @@ public void requestWithCookies() {
...
+			.expectBody()
+			.returnResult();
+		OperationRequest request = this.converter.convert(result);
+		assertThat(request.getCookies()).hasSize(1);
+		assertThat(request.getCookies()).extracting("name").containsExactly("sessionId");
+		assertThat(request.getCookies()).extracting("value").containsExactly("YWJjZGVm==");
+	}
+
+	@Test
+	void requestWithMalformedCookieHeaderWithoutEqualsSign() {
```
wilkinsona: This method has to be `public` as the 3.0.x branch uses JUnit 4.

wilkinsona (changes requested):

spring-restdocs-webtestclient/src/test/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverterTests.java
config25: Sorry about that, I missed the JUnit 4 requirement on the 3.0.x branch. Both test methods are now public. Thank you for catching this!

spring-restdocs-webtestclient/src/main/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverter.java
```diff
@@ -104,8 +104,11 @@ private Collection<RequestCookie> extractCookies(HttpHeaders headers) {
 	}
 
 	private RequestCookie createRequestCookie(String header) {
-		String[] components = header.split("=");
-		return new RequestCookie(components[0], components[1]);
+		int separator = header.indexOf('=');
+		if (separator == -1) {
+			return new RequestCookie(header, "");
+		}
```
wilkinsona: I don't think this is correct. To have an empty-string value, I believe the header's value would have to be `malformed=`. Without the equals sign, there should be no cookie at all as the header is invalid. I think failing in this case is fine so I'd just remove lines 108-110.

wilkinsona (changes requested):

spring-restdocs-webtestclient/src/main/java/org/springframework/restdocs/webtestclient/WebTestClientRequestConverter.java
config25: Thanks for the review! I've removed lines 108-110 as suggested.                                                                                                                                                                   
                                                                                                                                                                                                                                    
  Since the malformed cookie test (`requestWithMalformedCookieHeaderWithoutEqualsSign`)                                                                                                                                             
  was testing the exact behavior that's now removed, I've also deleted it — failing                                                                                                                                                 
  in that case is now the expected behavior, as you mentioned.

